PR #86, opened 2014-08-07
Fix Rails dependency

### LINKED ISSUE #74: Crashes when loading in a Sinatra app
When using in a Sinatra app, RailsConfig crashes because module Rails is not defined:

```
rails_config-0.4.1/lib/rails_config/engine.rb:2:in `<module:RailsConfig>': uninitialized constant Rails (NameError)
```

Suggest to only include engine.rb when Rails is defined.

adamcrown:
The current 0.4 version is dependent on Rails. This should remove the dependency on Rails::Engine and ActiveSupport so that Sinatra and other frameworks will be supported.

Fixes #74

### DIFF
--- a/lib/rails_config/options.rb
+++ b/lib/rails_config/options.rb
@@ -4,8 +4,8 @@ module RailsConfig
 
     include Enumerable
 
-    delegate :keys,   to: :marshal_dump
-    delegate :empty?, to: :marshal_dump
+    def keys() marshal_dump.keys; end
+    def empty?() marshal_dump.empty?; end
 
     def add_source!(source)
       # handle yaml file paths


jelmd:
Wondering, why there is still no fix available for a stable version. We just need to repeat ourselves, because < 0.5 is buggy and >= 0.5 is not stable ... :-((((

fredwu:
@jelmd You'd have to be more specific, what is not working for you? Please point to an existing issue, or create a new issue.

jelmd:
@fredwu: https://github.com/railsconfig/rails_config/pull/88 - 0.4.x still requires ::Rails

jelmd:
BTW: Are there any issues wrt. 0.5.0.beta1? If no, why not release it?

fredwu:
The latest gem version is v0.5.0.beta1, please use that and report issues if it doesn't work for you.

PR #89, opened 2014-08-29
Array descent with test

BCCRiskAdvisory:
This pull request includes the test created by frootloops to descend arrays when converting to a hash (updated to work with master as of 29-08-14), as well as the fix in the to_hash method to pass this test.


shanemcd:
Any plans on merging this in? Looks like there has been quite a few people that have had this issue. Confirmed the unexpected behavior still exists on master.

fredwu:
Thank you, merging! :)

PR #99, opened 2014-12-24
Change the default behavior for arrays from merging to overwriting

### LINKED ISSUE #12: Option overwriting doesn't work (for me, at least)
Hiya, 

I'm using an optional YML file with additional options as well as changes to the existing ones, which is loaded if it exists (doing so via `Settings.add_source!("#{Rails.root}/config/lpl.yml")` in my app's initializer). The problem is that I have the following defined in settings.yml:

    lpl: 
      ..
      language_options:
        - de
        - en
      ..

while my own file has

    lpl: 
      ..
      language_options:
        - de
      ..

I was expecting the result to be that `Settings.lpl.language_options` would hold `["de"]`, yet it still holds `["de", "en"]`. Do I need to do anything besides 

    if File.exists?("#{Rails.root}/config/lpl.yml")
      Settings.add_source!("#{Rails.root}/config/lpl.yml")
      Settings.reload!
    end

when loading the additional settings in my initlializer? Or should I be unloading all settings and include all the settings in `lpl.yml`, not just the ones that would add to/overwrite stuff from `config.yml`? And if so—how? 

Thanks! :)

hnakamur:
Hi, this is a pull request to fix #12.
Could you review this?
Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,46 @@ It won't work with arrays, though.
 
 To upload your local values to Heroku you could ran `bundle exec rake rails_config:heroku`.
 
+## Overwriting or Merging Arrays
+
+We changed the default behavior for arrays from merging to overwriting. Eg:
+
+settings.yml
+
+```yaml
+some_array:
+- 1
+- 2
+```
+
+settings.local.yml
+
+```yaml
+some_array:
+- 3
+- 4
+```
+
+config/initializers/rails_config.rb
+
+```ruby
+RailsConfig.setup do |config|
+  config.const_name = 'settings'
+end
+```
+
+`Settings.some_array` becomes `[3, 4]`.
+
+If you want the old behavior (merging arrays), just set the `overwrite_arrays` var to `false` in your `config/initializers/rails_config.rb` file. Eg:
+
+```ruby
+RailsConfig.setup do |config|
+  config.overwrite_arrays = false
+  config.const_name = 'settings'
+end
+```
+
+`Settings.some_array` becomes `[1, 2, 3, 4]`.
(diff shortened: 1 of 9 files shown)


vinceve:
+1 This looks good!

vinceve:
I tested it in our project. It doesn't work entirely. I noticed that my environment specific settings are overridden by the settings.yml. Normally it should work in the opposite way.

hnakamur:
@vinceve Thanks for testing. Could you share minimal settings.yml and settings.local.yml to reproduce problems?

vinceve:
My config:

settings.yml:
```yaml
service:
  hosts:
    - 'test'
```

settings/development.local.yml:
```yaml
service:
  hosts:
    - 'development'
```

if you do Settings.service.hosts:
=> [
  [0] "test"
]

hnakamur:
@vinceve I added more test cases and fixed merging with overwrite_arrays.
Could you try again?
Thanks.

vinceve:
@hnakamur I had sent you a merge request for your branch :-) Where I fixed it already yesterday. I'll check if it's the same.

hnakamur:
Sorry, I hadn't notice your pull request. My fix seems the same as yours.

fredwu:
Hi guys, great conversations!

As I commented in PR #103:

> Implementation wise, PR #110 has just been merged in - we now use the gem version of deep_merge, is this still achievable without changing deep_merge's source code?

pkuczynski:
Fixed in #137 although by default it does not overwrite, to maintain backward compatibility

PR #102, opened 2015-03-05
Add method prepend_source!

eugenk:
Sometimes you want to have custom configuration files that are loaded
before the default files, in order to let the `*.local.*` files
overwrite those settings. This method enables this behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,16 @@ Settings.reload!
 
 This will use the given source.yml file and use its settings to overwrite any previous ones.
 
+On the other hand, you can prepend a YML file to the list of configuration files:
+
+```ruby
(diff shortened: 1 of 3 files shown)


fredwu:
That's a great addition @eugenk! :+1: Could you please add some tests and add the usage to the README? I'm happy to merge this once tests and README are updated. :)

eugenk:
Sure. I just rebased on master and added the specs and usage.

eugenk:
The specs fail with ruby 1.9.3, but I can't imagine how the failed specs are linked to the new code of this pull request.

fredwu:
Nice work!

PR #103, opened 2015-03-11
Add the overwrite_arrays option

dtaniwaki:
I think we should be able to diable merging arrays.

e.g. Let's say you have the default settings like below.

`config/settings.yml`
```yaml
email:
  recipients:
    - foo1@example.com
    - foo2@example.com
```

And for production settings.

`config/settings/production.yml`
```yaml
email:
  recipients:
    - bar@example.com
```

`RailsConfig` merges the settings like below.

```yaml
email:
  recipients:
    - foo1@example.com
    - foo2@example.com
    - bar@example.com
```

However, I don't want to send email to `foo1@example.com` and `foo2@example.com` in production. Currently, we don't have any method to achieve this goal. So I just added an option to disable merging arrays.

What do you think this idea?

[Review on Reviewable](https://reviewable.io/reviews/railsconfig/rails_config/103)

### DIFF
--- a/lib/rails_config.rb
+++ b/lib/rails_config.rb
@@ -10,9 +10,10 @@ module RailsConfig
   # ensures the setup only gets run once
   @@_ran_once = false
 
-  mattr_accessor :const_name, :use_env
+  mattr_accessor :const_name, :use_env, :overwrite_arrays
   @@const_name = "Settings"
   @@use_env = false
+  @@overwrite_arrays = false
 
   def self.setup
     yield self if @@_ran_once == false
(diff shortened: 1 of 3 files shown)


fredwu:
Hi @dtaniwaki, this is a great idea! Implementation wise, PR #110 has just been merged in - we now use the gem version of `deep_merge`, is this still achievable without changing `deep_merge`'s source code?

dtaniwaki:
@fredwu 
Let me try it!

hnakamur:
@dtaniwaki Thanks for trying to update your pull request.

Maybe you can use test cases added by me. I updated my branch to rewrite test cases with rspec `expect`: https://github.com/railsconfig/rails_config/compare/master...hnakamur:overwrite_arrays

Note I have not updated my implementation. I leave it up to you.

pkuczynski:
Fixed in #137 using newer deep_merge gem

PR #104, opened 2015-03-11
Skip overwriting if the destination is nil

dtaniwaki:
I think we should not overwrite value if you purposely set nil.

e.g.

`config/settings.yml`

```yaml
foo:
  bar: 1
```

`config/settings/development.yml`

```yaml
foo:
  bar:
```

It should merge them like this.

```yaml
foo:
  bar:
```

instead of

```yaml
foo:
  bar: 1
```

What do you think?

[Review on Reviewable](https://reviewable.io/reviews/railsconfig/rails_config/104)


dtaniwaki:
It might break the settings if someone unconsciously leave a key without any value. It's a little risky so I close this.

Issue #106, opened 2015-04-01
NameError: uninitialized constant AppName::Application::Settings. Rails 4.2.1


Antiarchitect:
Same as #44 but.
rails_config 0.5.0.beta1
rails 4.2.1

fredwu:
Closing this as it should be fixed by #107.

PR #107, opened 2015-04-01
Allows to load RailsConfig manually

Antiarchitect:
Allows to load RailsConfig manually and even before app configuration process (e.g. inside config/application.rb):

```ruby
module Appname
  class Application < Rails::Application

    Bundler.require(*Rails.groups)
    RailsConfig::Integration::Rails::Railtie.preload

    ...

    config.time_zone = Settings.time_zone

    ...

   end
end
```

refers to #106

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,25 @@ Add this to your `Gemfile`:
 gem "rails_config"
 ```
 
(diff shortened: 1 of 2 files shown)


fredwu:
Neat!

PR #110, opened 2015-06-27
added gem deep_merge to gemspec

firedev:
When used in Rails together with Deep Merge gem Rails config produces the following error:

```
/[...]/gems/deep_merge-1.0.1/lib/deep_merge/core.rb:5: warning: already initialized constant DeepMerge::DEFAULT_FIELD_KNOCKOUT_PREFIX
/[...]/gems/rails_config-0.4.2/lib/rails_config/vendor/deep_merge.rb:4: warning: previous definition of DEFAULT_FIELD_KNOCKOUT_PREFIX was here
```

It is caused by special rails_compat module in deep_merge:
```
gem 'deep_merge', require: 'deep_merge/rails_compat'
```

That defines 'deeper_merge' methods on hash: https://github.com/danielsdeleo/deep_merge#using-deep_merge-in-rails

This pull request adds `gem deep_merge` as dependency to solve this once and for all.


fredwu:
This is great, thank you @firedev! :)

firedev:
Thanks, please bump the version up so I can use it in production without referring to GitHub.

Issue #108, opened 2015-04-28
Read environment variables


stefanahman:
Hi, I wonder if it is possible to do the following:
```bash
SETTINGS_I_LIKE_CATS="true" be rails c
```
and then recover the variable by:
```ruby
Settings.i_like_cats # => "true"
```

fredwu:
It is certainly possible, would you or anyone like to have a crack and submit a pull request? ;)

vinceve:
I would like to take a look at this. Wouldn't it be better to have some kind of prefix for env variables. For example:

```ruby
Settings.env.i_like_cats # => "true"
```

PR #119, opened 2015-10-03
Feature support environment variables

vinceve:
Support environment variables to use as settings. #108


pkuczynski:
I decided to use #125 as implementation of this feature requested in #108. Thank you for your work!

PR #125, opened 2016-01-04
Additional options for parsing env variables

spalladino:
Allows setting env vars like "SETTINGS__FOLDER__MY_KEY".

- env_prefix: change the prefix for the env vars to be processed
- env_separator: set the symbol or symbols on which to split
- env_converter: set to :downcase to change MY_KEY to my_key
- env_parse_values: convert integer-like env values to integers

This is similar to https://github.com/railsconfig/config/pull/119/, but builds on the existing `env_load!` method and provides further customisation on the format of the env variables. It should fix https://github.com/railsconfig/config/issues/108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,12 +263,9 @@ Inheritance customization (check [Deep Merge](https://github.com/danielsdeleo/de
 
 * `knockout_prefix` - ability to remove elements of the array set in earlier loaded settings file. Default: `nil`
 
+## Environment variables
 
-## Working with Heroku
-
-Heroku uses ENV object to store sensitive settings which are like the local files described above. You cannot upload such files to Heroku because it's ephemeral filesystem gets recreated from the git sources on each instance refresh.
-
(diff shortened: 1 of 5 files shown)


spalladino:
Please let me know if there is interest in this feature, in which case I'll rebase the changes on top of master so it can be merged.

cbeer:
This also fixes #125, I think?

spalladino:
@cbeer I think you mixed up the case ID

cbeer:
Fixed, thanks. This would be really useful for using e.g. AWS ElasticBeanstalk with config.

spalladino:
👍 Yes, it should fix #138 as well. We are running several of our apps as Docker containers in Dockercloud, and it is far easier to set env vars than to mount a config file.

pkuczynski:
This is a really good idea and implementation! I just merged it in and will prepare release 1.2.0. I also updated some documentation and added new options to Rails initializer

PR #129, opened 2016-01-26
Add gem name in Installation Instruction section

irfanfadilah:
Since there's no gem name in the documentation, especially in Installation section.

This also will help people who used this gem in the past (gem name back then was `rails-config`) in order to clear the confusion.


fredwu:
Good call! :+1:

irfanfadilah:
Thanks @fredwu.. :smile:

PR #130, opened 2016-02-06
Minor refactoring and cleanup

yhirano55:


pkuczynski:
And what would be the value of this change?

pkuczynski:
Rejecting this pull request due lack of response from the requestor.

Issue #120, opened 2015-11-10
DeepMerge - avoid conflict with active support


darrenleeweber:
See https://github.com/danielsdeleo/deep_merge#using-deep_merge-in-rails
```ruby
require 'deep_merge/rails_compat'
```

pkuczynski:
This issue is not valid in Config case, as we do not use to_hash method but directly call DeepMerge.merge!

PR #144, opened 2016-05-25
Merging environment variables should not override existing settings

cbeer:
We're having a problem with the new environment loading code and our configuration hierarchy.

With a setting like:

```yaml
feature:
  enabled: true
  group: x
```

If I have an environment variable (and config configuration to support..)  e.g. `SETTINGS_FEATURE_GROUP=y`, in the application I see: 

```ruby
Settings.feature.group # => 'y'
```

But, other keys in the configuration are removed:

```ruby
Settings.feature.enabled # => nil
```

It looks like there's a problem with string + symbol hashes coming out of `OpenStruct`. I don't know if we ought to assume OpenStruct will always dump symbolized hashes, or, as I do here, push the hash through `Config::Options` before dumping it.

### DIFF
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -107,7 +107,8 @@ module Config
 
     def merge!(hash)
       current = to_hash
-      DeepMerge.deep_merge!(hash.dup, current)
+      hash = Config::Options.new(hash.dup).to_hash
+      DeepMerge.deep_merge!(hash, current)
       marshal_load(__convert(current).marshal_dump)
       self
     end
--- a/spec/config_spec.rb
+++ b/spec/config_spec.rb
@@ -178,6 +178,14 @@ describe Config do
       expect(Settings.server).to eq("google.com")
       expect(Settings.size).to eq("3")
     end
+
+    it "should deep merge hashes" do
+      ENV['Settings.inner.something1'] = 'new blah'
+      config = Config.load_files("#{fixture_path}/deep_merge/config1.yml")
+
+      expect(config.inner.something1).to eq('new blah')
+      expect(config.inner.something2).to eq('blah2')
+    end
   end
 
   context "Nested Settings" do


lib/config/options.rb
```diff
@@ -83,17 +83,7 @@ def reload_from_files(*files)
...
-        if v.instance_of? Config::Options
-          result[k] = v.to_hash
-        elsif v.instance_of? Array
-          result[k] = descend_array(v)
-        else
-          result[k] = v
-        end
-      end
-      result
+      __convert_to_hash(marshal_dump)
```
pkuczynski: What would be the point of moving this to a separate method?

cbeer: Happy to move it back if you want. I initially thought we could use `__convert_to_hash` directly, but  ran into the symbol vs string problem.

pkuczynski: Well, this change brings no value and only makes the code more complicated. so yes, I would like it to be gone :) By the way, are you running this code in Rails? I have a suspicion it might be connected with #120

spec/fixtures/settings.yml
```diff
@@ -19,3 +19,7 @@ photo_sizes:
 root:
   yahoo.com: 2
   'google.com': 3
+
+feature:
+  enabled: true
+  group: x
```
pkuczynski: Instead of adding more sections to settings.yml, why don't we reuse what's already there?

lib/config/options.rb
cbeer: 👌 , removed.

Yes, I am running this code in Rails, but I'm not sure about the connection to DeepMerge. If I take the example from this spec and drop into a debugger at https://github.com/railsconfig/config/blob/master/lib/config/options.rb#L110:

```
[106, 115] in /Users/cabeer/Projects/3rd-party/config/lib/config/options.rb
   106:     end
   107: 
   108:     def merge!(hash)
   109:       require 'byebug'; byebug
   110:       current = to_hash
=> 111:       DeepMerge.deep_merge!(hash, current)
   112:       marshal_load(__convert(current).marshal_dump)
   113:       self
   114:     end
   115: 
```

The two hashes have different keys (strings vs symbols):
```
(byebug) hash
{
  [...]
  "size"=>3
  "inner"=>{
     "something1"=>"new blah"
  }
}
```

```
(byebug) current
{
  :size=>1,
  :inner=>{
    :something1=>"blah1",
    :something2=>"blah2"
   },
   [...]
  ...
}
```

And the merge (dutifully) merges them into an even bigger hash:

```
(byebug) current
{
  :size=>1, 
  :inner=>{:something1=>"blah1", :something2=>"blah2"},
  [...]
   "size"=>"3",
   "inner"=>{"something1"=>"new blah"}
}
```

With simple config structures (like `size` above), this isn't a problem when we load the hash back into OpenStruct; the last key (which happens to contain the data we want) wins.

With a complex structure (like `inner`), the last element replaces the value entirely, removing the previous settings.

pkuczynski:
I tested your branch and your fix did't really worked for me. So I spent quite some time last night and rebuild how ENV variables are loaded to correctly support multilevel settings. I just published the new version of the gem (1.2.1). Let me know if that fixes your problem?

cbeer:
Will do, thanks.

cbeer:
@pkuczynski 1.2.1 fixes the problem for me, thanks.

pkuczynski:
Great to hear that :)

Issue #145, opened 2016-05-28
In Readme it should be 'Config::Integrations::Rails::Railtie.preload'


altmer:
Currently in readme it says that in order to use config in initializers you should use code

```
Config::Integration::Rails::Railtie.preload
```

when actually it should be (with -s)

```
Config::Integrations::Rails::Railtie.preload
```

pkuczynski:
Fixed in #146 by @kris

pkuczynski:
@altmer, using `preload` mentioned in the README should not be required anymore, as Config should be loaded before Rails configuration. Example basic Rails app illustrating this case is presented on this repo: https://github.com/railsconfig/config-rails-4.2

I am writing to you, as I assume you used the README suggestion in you application. Could you please remove it from your app and check it that's still working? I assume it should and then I could remove this unnecessary description from README...

PR #146, opened 2016-05-29
Typo fix for Config::Integrations

kris:
The README is incorrect and needs to be updated. RE: #145

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ module Appname
   class Application < Rails::Application
 
     Bundler.require(*Rails.groups)
-    Config::Integration::Rails::Railtie.preload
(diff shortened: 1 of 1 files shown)


pkuczynski:
Merged. Thanks! :)

pkuczynski:
@kris, using `preload` mentioned in the README should not be required anymore, as Config should be loaded before Rails configuration. Example basic Rails app illustrating this case is presented on this repo: https://github.com/railsconfig/config-rails-4.2

I am writing to you, as I assume you used the README suggestion in you application. Could you please remove it from your app and check it that's still working? I assume it should and then I could remove this unnecessary description from README...

PR #137, opened 2016-04-01
Add overwrite_arrays option

Fryguy:
This is an extraction from
https://github.com/railsconfig/config/pull/103 where @dtaniwaki had
added this option, but had done it prior to the deep_merge extraction.

This depends on https://github.com/danielsdeleo/deep_merge/pull/21

Paired with @carbonin <https://github.com/carbonin> on this.

PR #147, opened 2016-07-27
Overwrite arrays

vinceve:
I took Fryguy's commit to add the overwrite arrays option and pushed the gem myself to rubygems. It looks like the deep_merge gem latest release was in 2011.

https://github.com/railsconfig/config/pull/137


Fryguy:
I've updated #137 now that a new version of deep_merge has been released.

pkuczynski:
Fixed in #137

PR #155, opened 2016-10-29
Add optional config validation (using dry-validation schema)

ok32:
Tests are present, docs are not. : )


pkuczynski:
We would need some docs and fix failing build, before we can accept this PR :)

ok32:
Here come the docs.
As for the build, I'm afraid there's bad news. It won't build with rubies older than 2.1. I didn't know about this limitation until I saw the failing build. This isn't good, is it?

PR #167, opened 2017-03-23
Add count to the reserved names list

carbonin:
We ran into this issue by using `:count` as a key in our settings [here](https://github.com/ManageIQ/manageiq/blob/542ae37e1c4c0a1917c0d3bd9fcd4e5f4864d19d/config/settings.yml#L1182)

/cc @yrudman @Fryguy

### DIFF
--- a/spec/fixtures/reserved_keywords.yml
+++ b/spec/fixtures/reserved_keywords.yml
@@ -1,2 +1,3 @@
 select: 123
 collect: 456
+count: 789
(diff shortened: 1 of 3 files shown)


spec/fixtures/reserved_keywords.yml
```diff
@@ -1,2 +1,4 @@
 select: 123
 collect: 456
+section:
```
pkuczynski: Why adding this keyword deeper in the tree instead of on the same level as other reserved keywords?

carbonin:
Sorry about that. Just fixed it up.

pkuczynski:
No worries. Looks good. Thanks! :)

PR #173, opened 2017-05-10
Add extra_sources option and allow to add hashes as a settings source

romanoji:
Hey! 
I've added new feature, which allows to define extra config sources before config load.
I think this feature will be useful, because starting from this version **config validation** feature will be available. Unfortunately, validation doesn't work well, when you want to add new config files - it's impossible to validate values that appear in the files added at the runtime (using `Settings.add_source!`).

This may be also a sort of remedy for railsconfig/config#142.

### DIFF
--- a/lib/config.rb
+++ b/lib/config.rb
@@ -33,14 +33,15 @@ module Config
     @@_ran_once = true
   end
 
-  # Create a populated Options instance from a settings file. If a second file is given, then the sections of that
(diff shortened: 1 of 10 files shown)


kamilpp:
👍

pkuczynski:
We could have a look at this idea @romanoji but you would have to solve conflicts first.

fxckyuo (approved):

PR #177, opened 2017-08-01
Allow a custom environment variable prefix to include the environment variable separator

rdodson41:
Currently, given the following configuration:
```
Config.setup do |config|
  config.use_env = true
  config.env_prefix = 'MY_CONFIG'
  config.env_separator = '_'
  config.env_converter = :downcase
end
```
an environment variable, such as `MY_CONFIG_USERNAME=rdodson41`, will be split into `['MY', 'CONFIG', 'USERNAME']`, and, therefore will fail to match against the prefix `MY_CONFIG`.

This pull request will allow a custom environment variable prefix to include the environment variable separator by splitting the prefix on the separator and, only then, comparing the beginning of the split variable name to the split prefix. As a result, `['MY', 'CONFIG', 'USERNAME']` will successfully match against the prefix `MY_CONFIG`, which would be split into `['MY', 'CONFIG']`.

I don't think that this is a breaking change, as currently, given such a configuration, no environment variables would be merged into the configuration, but I am not completely sure.

### DIFF
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -39,7 +39,9 @@ module Config
       ENV.each do |variable, value|
         keys = variable.to_s.split(Config.env_separator)
 
-        next if keys.shift != (Config.env_prefix || Config.const_name)
+        prefix = (Config.env_prefix || Config.const_name).to_s.split(Config.env_separator)
+
+        next if keys.shift(prefix.size) != prefix
 
         keys.map! { |key|
           case Config.env_converter
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -37,9 +37,10 @@ module Config
       hash = Hash.new
 
       ENV.each do |variable, value|
-        keys = variable.to_s.split(Config.env_separator)
+        separator = Config.env_separator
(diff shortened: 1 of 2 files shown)


pkuczynski:
Code climate complain about `Config::Options#split doesn't depend on instance state (maybe move it to another class?)`. At the same time `Config.env_prefix` and `Config.const_name` should be already strings, so no point to convert them.

Could you please inline split method when appropriate?

Other than that looks good! :)

rdodson41:
@pkuczynski Thanks, for the feedback. I reverted the commit in question. Should be good now.

rdodson41:
@pkuczynski I fixed the other two Code Climate issues; the single remaining issue does not originate in this pull request, so I left it unchanged.

With regard to `(Config.env_prefix || Config.const_name).to_s.split(separator)`, I agree that `.to_s` *should* not be necessary, as both of these configuration variables *should* be strings, but I retained `.to_s` so that this method does not raise an error if, for some reason, a user configures one or both variables as symbols.

pkuczynski:
Looks good! Thanks!

PR #179, opened 2017-09-03
Change Default Constant from Settings to RailsConfig

irfanfadilah:
Yesterday I encountered an issue when using Settings as constant name, I have a code where I required to include SendGrid module, that module also has class Settings. Then I tried to call a config somewhere:

```
include SendGrid
Settings.my_api_key
```

Raise:

```
NoMethodError: undefined method `my_api_key' for SendGrid::Settings:Class
```

So I changed my constant to RailsConfig and it's working fine. I don't know if this issue more related to the Ruby itself or not, but as for why made this PR because the name "Settings" is too general and might conflict with other modules/gems. The name represent gem name as well. Since this huge changes, I don't really mind if rejected, I just want to share my thought.  :v: 

Also it's might related to this issue: https://github.com/railsconfig/config/issues/126

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Config helps you easily manage environment specific settings in an easy and usab
 * config files support ERB
 * config files support inheritance and multiple environments
 * access config information via convenient object member notation
-* support for multi-level settings (`Settings.group.subgroup.setting`)
+* support for multi-level settings (`RailsConfig.group.subgroup.setting`)
 * local developer settings ignored when committing the code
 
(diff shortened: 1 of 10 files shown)


pkuczynski:
This PR is not needed. If you want to change the variable name, have a look at https://github.com/railsconfig/config#configuration - all you have to do is to change `config/initializers/config.rb` and add:

```ruby
Config.setup do |config|
  config.const_name = 'RailsConfig'
end
```

PR #180, opened 2017-10-19
Add parsing of ENV variable values to Boolean type

### LINKED ISSUE #178: Support booleans with env_parse_values
This is more of a feature request, but if you have a config like `something: true` and then override it with an ENV variable `SETTINGS__SOMETHING=false` the setting becomes "false" which is still truthy. I wonder if there's an eloquent way to make these values a boolean, or if that isn't an option, update the readme to explicitly call out not to use boolean configs as there will be no way to override them unless your code treats them like strings.

mjgiarlo:
Fixes #178 

(Discovered in https://github.com/samvera-labs/hyku/issues/1462. Heads-up, @tdonohue.)

### DIFF
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -188,7 +188,14 @@ module Config
 
     # Try to convert string to a correct type
     def __value(v)
-      Integer(v) rescue Float(v) rescue v
+      case v
+      when 'false'
+        false
+      when 'true'
+        true
+      else
+        Integer(v) rescue Float(v) rescue v
+      end
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### New features
+
+* Add parsing of ENV variable values to Boolean type
+
 ...
 
 ## 1.5.0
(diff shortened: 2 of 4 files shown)


mjgiarlo:
@pkuczynski I can eliminate the CodeClimate issue by taking the logic from `#__boolean` and inlining it in `#__value`, but I didn't want to change the one-liner pattern going there. What do you advise? Should I make CodeClimate happy, or are there other changes you'd like to see?

pkuczynski:
I would rather see this done as:

```ruby
# Try to convert string to a correct type
    def __value(v)
      case value
      when 'false'
        false
      when 'true'
        true
      else
        Integer(v) rescue Float(v) rescue v
      end
    end
```

lib/config/options.rb
```diff
@@ -188,7 +188,14 @@ def __convert(h) #:nodoc:
 
     # Try to convert string to a correct type
     def __value(v)
-      Integer(v) rescue Float(v) rescue v
+      case value
```
mjgiarlo: `s/value/v/` here, right?

pkuczynski: yeah, totally! sorry!

mjgiarlo:
@pkuczynski I fixed a typo in the change you submitted, squashed commits, and pushed again. Thanks for the change! I suspect this is now ready to go.

When do you think 1.6.0 can get released with this change?

pkuczynski:
Done! Released as `1.5.1`.

pkuczynski:
Thanks for your contributiuon :)

mjgiarlo:
Thanks for the useful gem, @pkuczynski!

PR #182, opened 2017-10-25
Adds the option `raise_on_access_to_missing` (default `false`)

### LINKED ISSUE #43: When did a missing setting started to return nil ?
Hi,

I hope this issue's title doesn't seem aggressive, anyway it's not the intention ;-)

For a long time, I've been checking for particular setting like this :

``` ruby
affiliate_id = begin
  Settings.tracking.affiliate_id.send(options[:aid])
rescue NoMethodError, TypeError, ArgumentError => ex
  Settings.tracking.affiliate_id.default
end
```

Today — I don't think my version has changed (0.2.x) — I get `nil` for a missing setting, instead of a `NoMethodError` exception.

I was surprised not to find any test case for this in the current `0.3.1`. I even wanted to add one, but i've had issues with Ruby 1.8 compatibility (hence my pull request and my #42 issue).

Am I crazy and it has always been that way?
Am I using it wrong?

Anyway, thanks for your help and for making/maintaining this great gem.

### LINKED ISSUE #77: Missing keys should raise an error
This really seemed like a perfect gem for me. I extactly wanted specification for different environments in separate files.

However, there is one major design flaw. Accessing an option that wasn't set in the YAML files should raise a `NoMethodError`. That is, `RailsConfig::Options` shouldn't be an `OpenStruct`. Otherwise there is no way to guard yourself from typos, since there is no way to determine whether you made one. And that can cause really hard-to-find bugs (e.g. boolean value like `Settings.use_secure_sever`).

### LINKED ISSUE #171: Raise on non-existent key
Another useful feature might be an option that instead of returning nil on a non-existent key, would instead raise an exception.  I feel this could be similar to Hash#fetch.

cc @cben

Yoshiji:
This option will raise a `KeyError: "key not found: #{key.inspect}"` when trying to access non-existing keys.
It also includes the `key?(name)` (and its alias`has_key?(name)` too) to test if a value was set for a given key.

```ruby
Config.setup { |cfg| cfg.fail_on_missing = true }
config = Config.load_files("empty.yml")
config.key?(:not_existing)
# => false
config.not_existing
# KeyError: key not found: :not_existing
```
Fixes #171, Fixes #77, Fixes #43

### DIFF
--- a/lib/config.rb
+++ b/lib/config.rb
@@ -15,13 +15,15 @@ module Config
   # Ensures the setup only gets run once
   @@_ran_once = false
 
-  mattr_accessor :const_name, :use_env, :env_prefix, :env_separator, :env_converter, :env_parse_values
+  mattr_accessor :const_name, :use_env, :env_prefix, :env_separator,
+                 :env_converter, :env_parse_values, :fail_on_missing
   @@const_name = 'Settings'
   @@use_env    = false
   @@env_prefix = @@const_name
   @@env_separator = '.'
   @@env_converter = :downcase
   @@env_parse_values = true
+  @@fail_on_missing = false
 
   # deep_merge options
   mattr_accessor :knockout_prefix, :overwrite_arrays
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### New features
 
-...
+* `Config#fail_on_missing` option (default `false`) to raise a `KeyError` exception when accessing a non-existing key
+* Add ability to test if a value was set for a given key with `key?` and `has_key?`
 
 ## 1.5.1
 
--- a/README.md
+++ b/README.md
@@ -299,6 +299,42 @@ between the schema and your config.
 
 Check [dry-validation](https://github.com/dry-rb/dry-validation) for more details.
 
+### Missing keys
+
+You can test if a value was set for a given key using `key?` and its alias `has_key?`:
+
+```ruby
+Settings.key?(:not_existing_key)
+# => false
+Settings.key?(:existing_key)
+# => true
+```
+
+By default, accessing to a missing key returns `nil`:
+
+```ruby
+Settings.key?(:not_existing_key)
+# => false
+Settings.not_existing_key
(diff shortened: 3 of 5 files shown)


pkuczynski:
Thanks for this PR! Looks great. Could you please:

* Rename `raise_on_access_to_missing` to `fail_on_missing`
* Update documentation (README, CHANGELOG)
* Fix Code Climate

Yoshiji:
See my amended commit.

lib/config.rb
```diff
@@ -15,13 +15,15 @@ module Config
...
-  mattr_accessor :const_name, :use_env, :env_prefix, :env_separator, :env_converter, :env_parse_values
+  mattr_accessor :const_name, :use_env, :env_prefix, :env_separator,
+                 :env_converter, :env_parse_values, :fail_on_missing
   @@const_name = 'Settings'
   @@use_env    = false
   @@env_prefix = @@const_name
   @@env_separator = '.'
   @@env_converter = :downcase
   @@env_parse_values = true
+  @fail_on_missing = false
```
pkuczynski: Shouldn't we use `@@` as for every other setting?

pkuczynski:
Added a comment. After we solve this, I can merge it and release a new version.

lib/config.rb
Yoshiji: I definitely agree, I changed it to `@` because Code Climate was complaining about it on my first version of this PR. If you are okay with using this class variable and make Code Climate upset about it, then I will rollback to `@@fail_on_missing`

pkuczynski: I think that’s better to keep consistency. Would you agree?

Yoshiji: I updated the code to keep consistency.

Yoshiji:
I updated the code to keep consistency. As foreseen, Code Climate and Codacy are upset but we decided to go this way.
Anything else I can do to improve this PR and/or code?

pkuczynski:
All good! Merged and I will release it shortly...

jcoyne:
Thanks! 👏

PR #186, opened 2017-12-10
Padrino should respond to env and root

tily:
## Summary

* In `lib/config/integrations/sinatra.rb`, `Padrino.env` and `Padrino.root` is called without `respond_to?` check
* But there are cases where only `padrino-helpers` is depended on, and `Padrino.env` and `Padrino.root` are not available, like `kaminari-sinatra`
    * See https://github.com/kaminari/kaminari-sinatra/blob/master/kaminari-sinatra.gemspec#L34


pkuczynski:
Thanks!

PR #190, opened 2018-01-10
Restrict constant search to Kernel only (#189)

### LINKED ISSUE #189: NameError for Kernel::Settings when the global const Settings is already defined.
This bug occurs in line 55 of lib/config.rb when the constant `Settings` is already defined, but somehow Kernel::Settings is not. I am not 100% clear on how to achieve this scenario. The culprit in my setup seems to be the configliere gem which is loaded by a dependency of my project.

The specific issue is that `const_defined?` has default behavior (`inherit=true`) that also searches Object.

malakai97:
This fixes #189 while retaining the warning suppression behavior.

### DIFF
--- a/lib/config.rb
+++ b/lib/config.rb
@@ -52,7 +52,7 @@ module Config
 
   # Loads and sets the settings constant!
   def self.load_and_set_settings(*files)
-    Kernel.send(:remove_const, Config.const_name) if Kernel.const_defined?(Config.const_name)
+    Kernel.send(:remove_const, Config.const_name) if Kernel.const_defined?(Config.const_name, false)
     Kernel.const_set(Config.const_name, Config.load_files(files))
   end
 
--- a/spec/const_loading_spec.rb
+++ b/spec/const_loading_spec.rb
@@ -1,0 +1,13 @@
+require "config"
+
+RSpec.describe Config do
+  let(:fixture_path) { File.expand_path('../fixtures', __FILE__) }
+  context "Object::Settings defined" do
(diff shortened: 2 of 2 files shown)


pkuczynski:
Looks good! Could you also provide test covering this issue?

malakai97:
Most likely yes, but I'm not sure when I'll get to that.

malakai97:
I can also fix the code climate line size issue by converting it to a standard if block, or we can leave it.

pkuczynski:
CodeClimate is easy to fix. I am more interested in a unit test covering this case. I hope you can work on this soon...

pkuczynski:
Any luck with adding some tests @malakai97?

malakai97:
I have created a test in `spec/const_loading_spec.rb`.  The reason I placed it here instead of just in `config_spec.rb` is that it passes without the fix if `spec_helper` has been loaded. I don't know why.

Thus, you can see the failure and subsequent pass via:

`bundle exec rspec spec/const_loading_spec.rb`

malakai97:
Possible strategies for getting it to fail when running the entire suite:

1. Figure out what spec_helper is doing that "fixes" it, and undo it in the test itself.
2. Don't include spec_helper by default, and instead require it in the spec files that need it.

I expect you don't want to do number two.

pkuczynski:
Indeed, number one is better :) I guess u would have to figure out how to emulate the scenario when the constant Settings is already defined, but somehow Kernel::Settings is not. Otherwise we can't be sure that propose change is doing what it should be.

PR #191, opened 2018-01-17
Update deep_merge dependency to latest version (v1.2.1)

tomkersten:
Minor version bump from 1.1.1->1.2.1. Removes `deep_merge` from
showing up as an outdated gem in `bundle outdated` output (of Rails 5 application).

All specs pass after the update on my system.


pkuczynski:
Thanks!

PR #197, opened 2018-02-02
Add `zip` to the list of reserved keywords

niclin:
When I call `Settings.some_setting.["zip"]` there always return Array


pkuczynski:
Thanks! Could you provide CHANGELOG entry and a unit test covering this case please?

niclin:
@pkuczynski done 😁

pkuczynski:
Good work, thanks! :)

PR #213, opened 2019-01-10
⚠ loading in progress, circular require considered harmful

amatsuda:
This `require` requires itself, then causes a ruby interpreter warning.
This can be reproduced by
```
% ruby -rconfig -wep
```
with current 1.7.1 gem.


pkuczynski:
Looks like your PR is failing the build

amatsuda:
Do you mean you want me to fix your build? Then here you are! #215

pkuczynski:
i believe this has been sorted now?

PR #215, opened 2019-01-10
Use bundler 1 on Travis CI for now

amatsuda:
Because we're still testing against ruby <2.3.0 that  is not supported by bundler 2 and rails 4.x that doesn't support bundler 2.x.


pkuczynski:
Thanks!

PR #216, opened 2019-02-10
Fix env_prefix default in README.

akira-kuriyama:
I think env_prefix default in README is wrong.
https://github.com/railsconfig/config/blob/2439d5a41061cf303aa912ce7f03b8b7e69ada4d/lib/config.rb#L22

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,7 +400,7 @@ To upload your local values to Heroku you could ran `bundle exec rake config:her
 
 You can customize how environment variables are processed:
 
-* `env_prefix` (default: `SETTINGS`) - which ENV variables to load into config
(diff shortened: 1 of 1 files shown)


pkuczynski:
That's true, but ENV variable is being expected in capitals AFAIR?

akira-kuriyama:
Yes, therefore I think the following is ideal.
`@@env_prefix = @@const_name.upcase`

In the current implementation,
```
Settings_KEY=value -> Settings.key == 'value'
SETTINGS_KEY=value -> Settings.key != 'value'
```
So, I think it would be better to modify either the document or the implementation.

PR #218, opened 2019-04-16
README.md: Update contributing instruction

bekicot:
The latest version of bundler doesn't support ruby 2.2.2, this gem is still supporting it.
This changes add instruction to install bundler that supports ruby 2.2.2.


pkuczynski:
Thanks!

PR #221, opened 2019-05-14
Fix broken tests

dsalahutdinov:


config.gemspec
```diff
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rspec',       '~> 3.7',   '>= 3.7.0'
   s.add_development_dependency 'rspec-rails', '~> 3.7',   '>= 3.7.2'
   s.add_development_dependency 'test-unit',   '~> 3.2',   '>= 3.2.7'
-  s.add_development_dependency 'sqlite3',     '~> 1.3',   '>= 1.3.13'
+  s.add_development_dependency 'sqlite3',     '~> 1.3.0',   '>= 1.3.6'
```
pkuczynski: I think we should keep `1.3` here

pkuczynski:
Could u also please fix 2 issues https://codeclimate.com/github/railsconfig/config/pull/221

gemfiles/rails_5.gemfile
```diff
@@ -2,6 +2,7 @@
 
 source "https://rubygems.org"
 
+gem "sqlite3", "< 1.4.0"
```
pkuczynski: Why would we need to lock sqlite everywhere?

dsalahutdinov: Hey @pkuczynski 
After release of sqlite3 gem version 1.4.0, tests failed because of 
```LoadError:
  Error loading the 'sqlite3' Active Record adapter. Missing a gem it depends on? can't activate sqlite3 (~> 1.3.6), already activated sqlite3-1.4.1. Make sure all dependencies are added to Gemfile.
```
It is happening because rails requires it [here|https://github.com/rails/rails/blob/5-2-stable/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L12]

pkuczynski: But that should happen only for newer version of rails, right?

pkuczynski: If so, we could do more logic on the `config.gemspec` as we do in some other cases.

dsalahutdinov: No, it works the same for every version of Rails (it's because rails formally does not depend on any db adapter, that is why it requires adapter gem at the runtime)

pkuczynski: Well, how did it worked before then? :)

PR #223, opened 2019-06-11
Limit max version of dry-validation 

### LINKED ISSUE #222: dry-validation 1.x moves schema support to dry-schema
Updating to dry-validation 1.x (which the gem spec allows) breaks config when `config.schema` is used with:

```
undefined method `Schema' for Dry::Validation:Module (NoMethodError)
```

It appears that the schema support has been moved to a new `dry-schema` gem so I guess some tweaking is needed to support that...

zbelzer:
Version 1.0 is a breaking change:
https://github.com/dry-rb/dry-validation/blob/master/CHANGELOG.md#v100-2019-06-10

Fixes #222


zbelzer:
OK, I'm pretty sure I can't make code climate happy since it thinks a line is too long, but if I put it on multiple it tells me it wants it on one again.

In lieu of any rewrites, can we please get a quick win on limiting the version?

spec/app/rails_5/config/initializers/new_framework_defaults.rb
```diff
@@ -17,8 +17,9 @@
 # Require `belongs_to` associations by default. Previous versions had false.
 Rails.application.config.active_record.belongs_to_required_by_default = true
 
+# Removed in rails 5.2+
```
pkuczynski: This was needed only for tests for Rails 5.0. There is a separate 5.2 folder. So this change should not be required?

lib/config/version.rb
```diff
@@ -1,3 +1,3 @@
 module Config
-  VERSION = '1.7.1'
+  VERSION = '1.7.2'.freeze
```
pkuczynski: You should not bump the version. `freeze` is OK

spec/app/rails_5/config/initializers/new_framework_defaults.rb
zbelzer: It blows up on `rake spec`. That must run the wrong combination of directories and versions? I had trouble getting appraisal to work here.

pkuczynski: Hmm, made some changes. Lets see if build fails.

lib/config/version.rb
zbelzer: I figured it needed to be bumped so bundler recognizes a new version when it's pushed. I'm used to doing it myself for work, forgot that might be part of a different process elsewhere.

spec/app/rails_5/config/initializers/new_framework_defaults.rb
zbelzer: Totally happy if it's not needed.

config.gemspec
```diff
@@ -25,11 +25,9 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'activesupport',  '>= 3.0'
   s.add_dependency 'deep_merge',     '~> 1.2',   '>= 1.2.1'
+  
   s.add_dependency 'dry-validation', '~> 0.12', '>= 0.12.2', '< 1.0.0' if RUBY_VERSION >= '2.2'
-
-  if RUBY_VERSION >= '2.1' && RUBY_VERSION < '2.2'
-    s.add_dependency 'dry-validation', '~> 0.10', '>= 0.10.7', '< 1.0.0'
-  end
+  s.add_dependency 'dry-validation', '~> 0.10', '>= 0.10.7', '< 1.0.0' if RUBY_VERSION >= '2.1' && RUBY_VERSION < '2.2'
```
zbelzer: Code climate complained that this line was too long. This is how I had it originally.

lib/config/version.rb
pkuczynski: I will upgrade this whenever the gem will get released. Saying that, I can release once this PR is merged I guess.

config.gemspec
pkuczynski: Does not seem to complain anymore?

pkuczynski:
Hmm it now fails in many places due sqlite issue. Would you have some time to debug?

zbelzer:
Sure, but it will probably be in the morning.

Appraisals
```diff
@@ -1,34 +1,42 @@
 appraise 'rails-3' do
+  gem 'sqlite3', "< 1.4.0"
```
pkuczynski: Could you use single quotes instead of double please?

pkuczynski:
We are getting very close! Still few things failing I am afraid :(

Appraisals
zbelzer: Whoops

zbelzer:
@pkuczynski Where are there always new code climate failures?

zbelzer:
OMGeee. I got the tests to pass. @pkuczynski Can we tell code climate to ignore those string macros or do you need me to add them.

pkuczynski:
Code climate fixed :)

zbelzer:
@pkuczynski all good?

pkuczynski:
Looks good to me! :) Should I release it? You also put a lot of effort into this PR - would you like to become a contributor?

zbelzer:
Please release it!

pkuczynski:
Already released :)

pkuczynski:
How about the contribution part?

zbelzer:
Sure. That would be great.

pkuczynski:
Done! :) Welcome onboard :)

PR #226, opened 2019-06-19
Add Ruby 2.7.0-preview1 to the test matrix and upgrade older versions

pkuczynski:


pkuczynski:
@rdubya would you like to help me with this one? I added upcoming ruby 2.7, and it's failing with error:

`undefined method `new_ostruct_member' for #<Config::Options>`

rdubya:
It looks like that method was deprecated in 2.4 and must have been removed in 2.7, not sure if I'll have time today but I might have time tonight to look at it

pkuczynski:
I have noticed that it has been deprecated and now removed, but not sure how to fix that. If you could have a look it would be great! Btw. should we already release 2.0 or wait for this one to be fixed?

rdubya:
Actually, it looks like method_missing might handle it by default now and the line can be removed. I pulled it out and the tests all passed on 2.4

eitoball:
I worked on this a bit. [This](https://github.com/eitoball/config/tree/upgrade-ruby-test-version) is my branch. This branch has successful [build](https://travis-ci.org/eitoball/config/builds/618104424).

Rails 6.0 had issue in ruby 2.7, and it has been fixed in rails/rails#37504.

pkuczynski:
Replaced by #255 and #264

PR #227, opened 2019-06-20
Define the constant on Object instead of Kernel

rdubya:
This addresses https://github.com/railsconfig/config/issues/122

It is branched off of the branch that upgrades the dry-rb gems so once that PR is merged, the changes here are minimal. Basically only config.rb and the changelog are updated.


pkuczynski:
Should this PR be treated as breaking change?

rdubya:
I doubt it will affect to many people but I think it should probably be considered a breaking change. I added it to the changelog as a breaking change, are there other things that need done to flag it that way as well?

pkuczynski:
I think I would not consider this as a breaking change if it does not affect people in a way they can't use this gem in certain conditions or a certain way.

PR #228, opened 2019-06-20
Add JRuby 9.2 to the test matrix

pkuczynski:


rdubya:
Would it make things easier if we pulled the bundler dependency out of the gemspec so we aren't forcing a specific version of bundler?

pkuczynski:
Yeah, good idea!

rdubya:
I realized the JRuby specs will fail because sqlite3 isn't compatible with jruby, we need to conditionally include activerecord-jdbcsqlite3-adapter in that case.

pkuczynski:
Do you wanna update this PR? Ideally, we should get rid of ActiveRecord dependency. It doesn't really make sense for this gem, don't you think?

pkuczynski:
If I understand correctly https://stackoverflow.com/questions/185573/what-is-mattr-accessor-in-a-rails-module the `attribute_accessors` are only needed to provide `mattr_accessor`, so maybe we can just replace it with explicit setters/getters definition? What do you think?

.travis.yml
```diff
@@ -22,7 +22,7 @@ matrix:
       gemfile: gemfiles/rails_4.2.gemfile
 before_install:
  # Bundler 2.x is supported for Rails > 4.2
- - ([[ "$BUNDLE_GEMFILE" != *"rails_4.2.gemfile" ]] && gem uninstall bundler -v '>= 2' && gem install bundler -v '< 2' ) || true
+ - ([[ "$BUNDLE_GEMFILE" = "$PWD/gemfiles/rails_4.2.gemfile" ]] && gem uninstall bundler -v '>= 2' && gem install bundler -v '< 2' ) || true
```
pkuczynski: How about `[[ "$BUNDLE_GEMFILE" =~ rails_4.2.gemfile$ ]]`

rdubya: That works for me, for some reason JRuby still doesn't like it even with this running with 4.2 so I gave up on it for the day. I can make that change the next time I get back into it.

pkuczynski: I applied it now. Let's see if it makes any difference. Travis is making me really tired. Builds take ages and cant install rbx. Strange. Maybe we should give a try to CircleCI?

pkuczynski:
@eregon seems that your jruby is built without or with wrong libyaml: https://github.com/rubyconfig/config/pull/228/checks?check_run_id=380015086

Can you help?

eregon:
It looks like a bug of JRuby when running in GitHub Actions.
Not sure what's causing it.
I found this issue: https://github.com/bundler/bundler/issues/6878#issuecomment-572205353
But it would good to file a new issue at https://github.com/jruby/jruby

eregon:
JRuby is not using libyaml AFAIK, they have their own version of SnakeYAML AFAIK.

pkuczynski:
Hmm, JRuby seems to be very flaky. Worked for https://github.com/rubyconfig/config/pull/228/commits/8ae80e94e54139eedb55ba7df8498785e5fdc385 and failed for https://github.com/rubyconfig/config/pull/228/commits/74b650fd3d21a595c10522bf14e64703de2963a5

This is really strange...

eregon:
I filed https://github.com/jruby/jruby/issues/6023

headius:
> Hmm, JRuby seems to be very flaky.

I assume you mean JRuby support on Github actions is flaky. If JRuby works once it should work every time. I strongly suspect there's some change in the Github actions environment that's messing with how we locate the Psych library, like it's picking up a GEM_HOME pointing at CRuby gems.

headius:
Also confirming @eregon's statement: JRuby does not use libyaml; we use SnakeYAML, which is shipped in-the-box with our psych extension as part of the JRuby distribution.

Literally any system with a working JDK should be able to run JRuby from that tarball with no additional dependencies or build steps. This is an environment issue.

pkuczynski:
@headius yeah, that's what I mean :)

I also posted an example on https://github.com/jruby/jruby/issues/6023

If you have any suggestions how can we get this build working, I am very keen to hear about it! Maybe it's connected with the gem cache from GHA?

headius:
> Maybe it's connected with the gem cache from GHA?

I think you nailed it. I just noticed that the failed build was able to restore a cache and the successful one was not. I suspect it's assuming the CRuby cached gems are ok for JRuby and loading the CRuby psych somewhere in the pipeline.

pkuczynski:
That's odd as I include `jruby` in my cache key, so it should be only relevant to jruby...

pkuczynski:
@headius it can not be a cache issue. Now the build had no-cache (I intentionally changed the key to test it) and it still failed the same way: https://github.com/rubyconfig/config/runs/380132707

headius:
@pkuczynski See https://github.com/jruby/jruby/issues/6023#issuecomment-572238722

This is a horrible thing to have set in a default CLASSPATH. GHA needs to remove that. For now I'd suggest clearing that var.

pkuczynski:
Waiting for https://github.com/actions/virtual-environments/issues/242 to be solved

PR #231, opened 2019-06-28
Execute tests in random order

### LINKED ISSUE #198: Specs rely on test order
## Description

Running the specs in a random order results in multiple failed specs, depending on the order's seed.  

## Rubies 

* 2.3.4
* 2.4.2 - appraisal will not run the entire suite regardless of errors encountered

## Steps to Reproduce

1. git clone, etc
2. `bundle install`
3. `appraisal install`
4. add `--order random` to the `.rspec` file
5. `appraisal rspec`

### Seeds which are known to result in errors

* 19209 - Multiple KeyErrors
* 5710 - Expectation failures in config_env_spec
* 61592 - Exactly one failed test in setting the constant to use for settings

pkuczynski:
Fixes #198


pkuczynski:
@pyromaniac I thought your last PR would fix this issue (by moving to instance variables). Do you have any idea why its still not the case?

pkuczynski:
@supremebeing7 maybe you want to have a quick look?

supremebeing7:
Nothing jumps out to me, I'd have to spend more time diving in and I don't have that time available right now. I will check it out when I do, if it's still an issue by then.

pkuczynski:
@cjlarose do you wanna look into this one too? It's a pretty poor setup we have, that test needs to be executed in a certain order. Would be good to clean them up, make them independent, and maybe even limit the number of fixtures we have to use...

cjlarose:
Yeah, I can take a look. It'd be good to get 'em running in random order and to prevent new tests from getting in that depend on the execution order.

PR #233, opened 2019-07-08
Ignore .local files in test environment

### LINKED ISSUE #135: Should test environment ignore .local files?
I'd argue yes.

Since you cannot know what other developers have configured in their `*.local.yml` files, you don't know how those values will affect the tests.
But we want tests to be reproducible, so I'd argue that in the `test` environment, these files should not be loaded. (After all, the behaviour can be changed using `reload_from_files` in test setup.)

What do you think?

franzliedke:
Without an option - was that requested or not? What should that look like?

Fixes #135.

### DIFF
--- a/lib/config.rb
+++ b/lib/config.rb
@@ -58,11 +58,16 @@ module Config
       File.join(config_root, 'settings.yml').to_s,
       File.join(config_root, 'settings', "#{env}.yml").to_s,
       File.join(config_root, 'environments', "#{env}.yml").to_s,
+      *local_setting_files(config_root, env)
+    ].freeze
+  end
 
-      File.join(config_root, 'settings.local.yml').to_s,
+  def self.local_setting_files(config_root, env)
+    [
+      (File.join(config_root, 'settings.local.yml').to_s if env != 'test'),
       File.join(config_root, 'settings', "#{env}.local.yml").to_s,
       File.join(config_root, 'environments', "#{env}.local.yml").to_s
-    ].freeze
+    ].compact
   end
 
   def self.reload!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changes
 
 * Get rid of activesupport dependency ([#230](https://github.com/railsconfig/config/pull/230))
+* Ignore .local files in test environment ([#135](https://github.com/railsconfig/config/issues/135), [#233](https://github.com/railsconfig/config/pull/233))
 
 ## 2.0.0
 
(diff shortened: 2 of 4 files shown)


pkuczynski:
@pyromaniac and @rdubya what do you think? Should we ignore local files in test env?

franzliedke:
Those truffleruby failures seem unrelated.

lib/config.rb
```diff
@@ -57,12 +57,14 @@ def self.setting_files(config_root, env)
...
-      File.join(config_root, 'environments', "#{env}.yml").to_s,
-
-      File.join(config_root, 'settings.local.yml').to_s,
-      File.join(config_root, 'settings', "#{env}.local.yml").to_s,
-      File.join(config_root, 'environments', "#{env}.local.yml").to_s
-    ].freeze
+      File.join(config_root, 'environments', "#{env}.yml").to_s
+    ].tap do |files|
+      files.concat([
+        File.join(config_root, 'settings.local.yml').to_s,
```
rdubya: It seems like this is the only file that should be affected by this. The next two lines would only include `settings/test.local.yml` and `environments/test.local.yml` which seem like they should still be included if somebody took the time to create them.

lib/config.rb
```diff
@@ -57,12 +57,14 @@ def self.setting_files(config_root, env)
     [
       File.join(config_root, 'settings.yml').to_s,
       File.join(config_root, 'settings', "#{env}.yml").to_s,
-      File.join(config_root, 'environments', "#{env}.yml").to_s,
-
-      File.join(config_root, 'settings.local.yml').to_s,
-      File.join(config_root, 'settings', "#{env}.local.yml").to_s,
-      File.join(config_root, 'environments', "#{env}.local.yml").to_s
-    ].freeze
```
rdubya: Can you make sure that we still end up with a frozen array? I'm not sure if it is a big deal or not, but it might bite somebody if we stop freezing it.

franzliedke: Woah, didn't mean to remove that. Good catch, fixed.

lib/config.rb
franzliedke: Hmm, true. I don't quite see why somebody should do that, but if they add a file specifically for the test environment, they probably want it to be included. :grinning:

franzliedke: Done.

pkuczynski:
Missing changelog entry, readme update, and codeclimate failures...

franzliedke:
Done.

pkuczynski:
Looks good! Thanks! :)

pkuczynski:
This gem needs a bit more love. @franzliedke maybe you would like to become a contributor?

franzliedke:
@pkuczynski Thanks for reaching out! I think I have to decline, though - busy enough keeping up with my own open-source projects, I'm afraid. :confused: 

Thanks for maintaining this and putting your work out there!

pkuczynski:
Sure, no worries! :) I know how it feels to be overwhelmed with other projets :)

PR #238, opened 2019-07-16
Add dry-validation contract support

### LINKED ISSUE #237: Update to use new dry-validation contracts
The dry-validation gem now relies on a new object called a `Contract`: https://dry-rb.org/gems/dry-validation/

These objects incorporate `schemas` and also allow you access to `rules` which make it possible to do conditional validation (e.g. "this should be a string if this other setting is odd").

**In theory**, this should be easy to implement by offering something like:

```ruby
class ConfigContract < Dry::Validation::Contract
  params do
    required(:email).filled(:string)
    required(:age).value(:integer)
  end

  rule(:email) do
    unless /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i.match?(value)
      key.failure('has invalid format')
    end
  end

  rule(:age) do
    key.failure('must be greater than 18') if value < 18
  end
end

Config.setup do |config|
  config.dry_validation_contract = ConfigContract
end
```

ryanwjackson:
Fix #237 

@pkuczynski Please let me know if there is anything missing or you would like changed.


rdubya:
👍 looks ok to me

README.md
```diff
@@ -7,7 +7,8 @@
 
 ## Summary
 
-Config helps you easily manage environment specific settings in an easy and usable manner.
+Config helps you easily manage environment specific settings in an easy and
```
pkuczynski: Could you please revert all those new line changes from your editor? We do not use hard wrapping...

ryanwjackson: @pkuczynski Done.  I was just following the linter errors.  May want to disable that one if you don't expect it to be fixed?

ryanwjackson:
@pkuczynski @rdubya which of these do you want me to address? https://codeclimate.com/github/railsconfig/config/pull/238#

For the `nil` check, any recommendations on how you would like this fixed?

lib/config.rb
```diff
@@ -23,7 +24,8 @@ module Config
     knockout_prefix: nil,
     merge_nil_values: true,
     overwrite_arrays: true,
-    merge_hash_arrays: false
+    merge_hash_arrays: false,
+    validation_contract: nil
```
pkuczynski: I am still wondering if we should call it `validation_contract` or simply `contract`. @pyromaniac @rdubya @supremebeing7

pkuczynski:
@ryanwjackson would https://github.com/railsconfig/config/pull/238/commits/67624d41e8592f4f9f043b3f1545e9f9385b134f makes sense to you?

lib/config.rb
supremebeing7: I'm in favor of being explicit - `validation_contract`

pkuczynski:
Looks good to me. Waiting for tests to pass and I will merge it in!

PR #246, opened 2019-08-06
Fix inconsistent documentation for ENV prefix and default value in generator

### LINKED ISSUE #162: Inconsistent documentation for ENV prefix
I believe these are the correct settings for bash-style env vars:

```
# Initializes the config gem
Config.setup do |config|
  config.use_env = true
  config.env_prefix = 'SETTINGS'
  config.env_separator = '__'
  config.env_converter = :downcase
end
```

Note that env_prefix needs to be all uppercase. What is there now is `Settings`, which doesn't work for me.

Thanks for such a great gem!

pkuczynski:
Fixes #162


smoyth:
LGTM, thanks!

pkuczynski:
Your welcome! Sorry for the late reply!

PR #247, opened 2019-08-16
Config doesn't have to be a Rails Engine

amatsuda:
I found that Config defines an isolated Engine class under Rails environment, but this simple plugin actually does not use Rails Engine feature at all. This plugin already has a Railtie, and that's enough for doing things that this plugin offers.

Creating an unused Engine class like this is not just unnecessary but indeed rather harmful for the users' applications, because it slows down the application initialization process, consumes much memory, and slows down auto/eager loading, request handling, rendering, etc. by adding entries for them.


pkuczynski:
@amatsuda thank you for your contribution. Could you please add CHANGELOG entry?

@pyromaniac @rdubya @supremebeing7 what you think about this PR?

amatsuda:
@pkuczynski 
> Could you please add CHANGELOG entry?

Done!

pkuczynski:
Thanks! Sorry for such a long time for merging this one...

pkuczynski:
Will be released soon with 2.2.0

PR #250, opened 2019-10-16
Defer modification of ActionController::Base to when it is loaded

cjlarose:
The `railtie` for this gem adds an initializer that loads the class `ActionController::Base` in order to add a `before_action` to it.

The initializer references the `ActionController::Base` constant directly, which, if it isn't already `load`ed, will be loaded at that moment. `load`ing that class causes other initializers run, including, for example, the [initializer defined by `actiontext`'s engine][0], as well as various initializers defined by third-party gems in a given application.

[0]: https://github.com/rails/rails/blob/65bc507973fe7105dc814f08f19fcc156c7daa10/actiontext/lib/action_text/engine.rb#L37

The effect is that adding the `config` gem to a Rails application that didn't have it causes that application to load `ActionController::Base` (and possibly a bunch of related stuff) during initialization, even in contexts where usage of `ActionController::Base` isn't expected, like when launching a console through `rails console`.

The solution here is to use `ActiveSupport::LazyLoadHooks` to add a hook that runs only after the `ActionController::Base` has been loaded.


rdubya:
Seems like a good idea to me, but can you look into the failing travis build and see if it is related to your change? It doesn't seem like it would be, but I don't see those failures in the last couple of commits to master.

cjlarose:
I'll take a look and add the changelog entry. Thanks!

cjlarose:
The test failures ended up being unrelated to my change, but I fixed it anyway.

The [most recent build on `master`](https://travis-ci.org/railsconfig/config/jobs/585154548#L296) happened to install `sprockets` version 3.7.2 when installing the sample Rails 4.2 application. But when [my PR ran a month later](https://travis-ci.org/railsconfig/config/jobs/598867733#L267), `sprockets` had published a new major version (4.0.0). For whatever reason, the sample Rails 4.2 application in this project is incompatible with sprockets 4.0.0 (not surprising, considering it's a major version release). All I did was modify the `gemfile` for the Rails 4.2 application so that it specifies that we want `sprockets` 3.7.x.

rdubya:
Awesome, thanks for finding/fixing the sprockets issue.

jrafanie:
@cjlarose I didn't see your fix for tests with rails 4.2 with sprockets 4 so I extracted a single PR to do it,  #254

lib/config/integrations/rails/railtie.rb
```diff
@@ -23,7 +23,7 @@ def preload
         # Development environment should reload settings on every request
         if ::Rails.env.development?
           initializer :config_reload_on_development do
-            ActionController::Base.class_eval do
+            ActiveSupport.on_load :action_controller_base do
```
pkuczynski: Can you dbl check which version is correct: yours or https://github.com/railsconfig/config/pull/258/files#diff-e1b40370ec751eed3246e103922adc0eR26 ?

pkuczynski: @amatsuda @Fryguy which version is better? This one or the one from #258?

Fryguy: I think this version is better.  I don't think the class_eval is needed because on_load takes care of it being in the right context.

pkuczynski: Yeah, that's what I thought too! Then I will merge this one!

pkuczynski:
Thank you @cjlarose!

PR #251, opened 2019-10-19
Drop support for Rails 4.2

cjlarose:
Rails 4.2 is [no longer officially maintained][0]. This PR updates the README to make clear that Rails 4.2 is not supported by `config` and also drops Rails 4.2 from the test suite.

[0]: https://guides.rubyonrails.org/maintenance_policy.html


pkuczynski:
The current version of the gem still works well with Rails 4.2, so I am not sure if its worth going hardcore and dropping support for it? #256 should fix failing tests. Rests should work, right?

cjlarose:
Yeah, if it's not a burden to keep support for Rails 4.2 in the codebase, then there's not a really compelling reason to drop support. That said, Rails 4.x users can always just install an older version.

rdubya:
We still use this with a 4.2 app, so I'd vote for keeping it around unless it becomes too much of a pain to deal with.

pkuczynski:
Rails 4.2 does not seem to require any extra changes in the code and everything seems to work smoothly, so there is no reason to drop it and cut people off from latest improvements.

Thanks for your PR anyway!

PR #253, opened 2019-11-12
Require dry-validation only when schema is specified

jrafanie:
If you never specify a schema, you're still paying the penalty of loading a
rather large library.

By waiting until someone specifies a schema to validate, we can reduce the
config require time and object allocations by ~30% if you don't specify an
optional schema. On my mac, bundle console is 30 MB before this change and
around 23 MB after.

**Before:**
$ time be ruby -e "require 'config'; puts GC.stat[:total_allocated_objects]"
347010
bundle exec ruby -e "require 'config'; puts GC.stat[:total_allocated_objects]  0.54s user 0.17s system 98% cpu 0.722 total

**After:**
$ time be ruby -e "require 'config'; puts GC.stat[:total_allocated_objects]"
224764
bundle exec ruby -e "require 'config'; puts GC.stat[:total_allocated_objects]  0.38s user 0.12s system 98% cpu 0.508 total


jrafanie:
The rails 4.2 test failures are fixed in #254

pkuczynski:
Looks great! Thanks!

PR #254, opened 2019-11-13
Pin to sprockets 3 as 4.0 support will require some work (fix tests with rails 4.2)

jrafanie:
See: https://github.com/rails/sprockets-rails/issues/443
and https://github.com/rails/sprockets/blob/master/UPGRADING.md#guide-to-upgrading-from-sprockets-3x-to-4

Note, config gem currently supports rails 4.2+ so it's unclear how easy it will
be to support sprockets 4 and rails 4.2.  For now, we can use sprockets 3 and
have tests pass.

### DIFF
--- a/config.gemspec
+++ b/config.gemspec
@@ -36,6 +36,11 @@ Please consider donating to our open collective to help us maintain this project
   s.add_development_dependency 'appraisal', '~> 2.2', '>= 2.2.0'
   s.add_development_dependency 'rspec', '~> 3.7', '>= 3.7.0'
 
+  # Sprockets 4.0 support will require some work:
+  # See: https://github.com/rails/sprockets-rails/issues/443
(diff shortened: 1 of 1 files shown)


cjlarose:
#250 is still open an includes the same change. Alternatively, #251 drops support for Rails 4.2 and would obviate the need for this change.

jrafanie:
@cjlarose yeah, I'll leave this open and @pkuczynski can offer opinions:
* drop rails 4.2 would probably need a version bump (#251), I can then close this PR
* #250 includes the same change as this PR so if that gets merged, I can close this one

I'm totally fine merging #250 and I can close this.  Ultimately, I wanted to get #253 in and didn't want to include this change in it.

pkuczynski:
I think the way it was done in #250 is better as it only affects Rails 4.2

pkuczynski:
I think the way it was done in #256 is better as it only affects Rails 4.2

PR #255, opened 2019-11-29
Fix missing new_ostruct_member in Ruby 2.7

qnighy:
`OpenStruct#new_ostruct_member` has been removed in https://github.com/ruby/ruby/pull/2178.


jcoyne:
Ref #259

lib/config/options.rb
```diff
@@ -201,15 +201,19 @@ def __convert(h) #:nodoc:
...
-        s.new_ostruct_member(k)
 
         if v.is_a?(Hash)
           v = v["type"] == "hash" ? v["contents"] : __convert(v)
         elsif v.is_a?(Array)
           v = v.collect { |e| e.instance_of?(Hash) ? __convert(e) : e }
         end
 
-        s.send("#{k}=".to_sym, v)
+        if s.respond_to?(:[]=)
```
jcoyne: Can you add a comment about why this conditional is needed?

qnighy: This was meant for backward compatibility with older Ruby versions. Now I noticed that the version `#[]=` was introduced is way older than the version `config` claims to support... https://github.com/ruby/ruby/commit/e44e356b53828d6468114b30e5c169296896f9b1#diff-f0ac5119f74bfda273347c9925e910d7

Therefore I'm going to remove this conditional, rather than adding comments.

Fryguy:
@qnighy Does it make sense to add 2.7.0 to the .travis.yml as well to verify this is working?

jrafanie:
Note, the failed tests are fixed in the hardcoding of sprockets in #250.

qnighy:
@Fryguy 

> @qnighy Does it make sense to add 2.7.0 to the .travis.yml as well to verify this is working?

Done!

jrafanie:
Looks like the rails 4.2 test suite doesn't play nice with ruby 2.7 and the appraisal version it's trying to use.  Is it time to drop rails 4.2 support?  We've had several PRs that all seem to have issues with 4.2 and it's no longer supported upstream: see these PRs that hit 4.2 issues: #250 #251 #253 #254.   @pkuczynski how can we help?

Fryguy:
You can probably also add an exclude to .travis.yml for 

```
    - rvm: 2.7.0
      gemfile: gemfiles/rails_4.2.gemfile
```

pkuczynski:
Thanks!

PR #258, opened 2019-12-10
Add Rails 6.0 to the test matrix

ojab:
prevents deprecation warning with rails-6: `DEPRECATION WARNING: Initialization autoloaded the constants ActionText::ContentHelper and ActionText::TagHelper`.

Also while we're here add rails-6 to the travis matrix


ojab:
Also fixed CI, so it's green.

ojab:
I can break it into separate commits/PRs (CI fix, rails-6 support, fixing deprecation) if you prefer.

spec/app/rails_6.0/app/assets/config/manifest.js
```diff
@@ -0,0 +1,2 @@
+//= link_tree ../images
```
pyromaniac: Can we generate a bare minimum rails app? The API one indeally

ojab: Sure, done.

config.gemspec
```diff
@@ -37,7 +37,7 @@ Please consider donating to our open collective to help us maintain this project
   s.add_development_dependency 'rspec', '~> 3.7', '>= 3.7.0'
 
   # Default RSpec run will test against latest Rails app
-  unless ENV['APPRAISAL_INITIALIZED']
+  unless ENV['APPRAISAL_INITIALIZED'] || ENV['CI']
```
pkuczynski: I think `ENV['APPRAISAL_INITIALIZED']` should be enough, don't you think?

ojab: /shrug I thought so, but it was failing on travis before this change

lib/config/integrations/rails/railtie.rb
```diff
@@ -23,11 +23,13 @@ def preload
         # Development environment should reload settings on every request
         if ::Rails.env.development?
           initializer :config_reload_on_development do
-            ActionController::Base.class_eval do
-              if ::Rails::VERSION::MAJOR >= 4
-                prepend_before_action { ::Config.reload! }
-              else
-                prepend_before_filter { ::Config.reload! }
+            ActiveSupport.on_load :action_controller do
```
pkuczynski: Wouldnt https://github.com/railsconfig/config/pull/250/files#diff-e1b40370ec751eed3246e103922adc0eR26 be enough (without `ActionController::Base.class_eval`)?

.travis.yml
```diff
@@ -23,6 +24,14 @@ matrix:
       gemfile: gemfiles/rails_5.1.gemfile
     - rvm: truffleruby
       gemfile: gemfiles/rails_5.2.gemfile
+    - rvm: truffleruby
+      gemfile: gemfiles/rails_6.0.gemfile
+    # rails-4.2 does not support ruby-2.6
```
pkuczynski: Don't think this is needed. Let me try to restore the previous state and see if the build passes?

PR #260, opened 2019-12-27
Fix validation contract documentation

tsubery:


pkuczynski:
@tsubery could you explain why this change is needed?

tsubery:
The documentation is wrong and i got an exception when i followed it. It works when i instantiate the object like this [spec](https://github.com/railsconfig/config/blob/fc57ca6c4004ae9d4ad8e951799995beb3fde8f4/spec/validation_spec.rb#L22)

pkuczynski:
Cool, thanks for the fix!

PR #261, opened 2019-12-29
Add support for ruby 2.7

jcoyne:
new_ostruct_member is no longer a method in ruby 2.7


qnighy:
#255 contains an approach that doesn't rely on private methods.

jcoyne:
@qnighy I think that #255 is probably better, but I saw the build was broken on that. Not until I tried this approach did I realize the master build is broken.

jcoyne:
Prefer #255 over this.

PR #264, opened 2020-01-04
Upgrade development dependencies and test matrix with latest Ruby and Rails versions

pkuczynski:


config.gemspec
```diff
@@ -28,7 +28,7 @@ Please consider donating to our open collective to help us maintain this project
   s.required_ruby_version = '>= 2.4.0'
 
   s.add_dependency 'deep_merge', '~> 1.2', '>= 1.2.1'
-  s.add_dependency 'dry-validation', '~> 1.0'
+  s.add_dependency 'dry-validation', '~> 1.4.0'
```
jcoyne: Why are you pinning so tightly here? A comment would be helpful.

config.gemspec
```diff
@@ -47,10 +47,10 @@ Please consider donating to our open collective to help us maintain this project
   s.add_development_dependency 'mdl', '~> 0.8', '>= 0.8.0'
 
   # Version 0.62 requires Ruby 2.2
```
jcoyne: Remove?

config.gemspec
pkuczynski: Do you think `s.add_dependency 'dry-validation', '~> 1.4'` would be better?

config.gemspec
pkuczynski: Indeed :)

config.gemspec
jcoyne: Yes, much better. Thanks

PR #272, opened 2020-02-10
Add hanami support

ippachi:
I add hanami support. 🌸
```
# in Gemfile
group :plugins do
  gem "config"
end
```

```
bundle exec hanami config
``` 
This code works well in `1.7.0` of rubyconfig/config.
But this is not usable in `2.2.1`. Because `dry-validation` version is difference in gemspec

```
# in rubyconfig/config
s.add_dependency 'dry-validation', '~> 1.0', '>= 1.0.0'	

# in hanami/validations	
spec.add_dependency 'dry-validation', '~> 0.11', '< 0.12'
```

I don't know what to do for this problem.
I hope this PR helps you!. 😄

### DIFF
--- a/lib/config.rb
+++ b/lib/config.rb
@@ -80,3 +80,7 @@ require('config/integrations/rails/railtie') if defined?(::Rails)
 
 # Sinatra integration
 require('config/integrations/sinatra') if defined?(::Sinatra)
+
+# Hanami integration and generator
+require('generators/config/hanami_install_generator') if defined?(::Hanami)
+require('config/integrations/hanami') if defined?(::Hanami)
--- a/lib/config/integrations/hanami.rb
+++ b/lib/config/integrations/hanami.rb
@@ -1,0 +1,9 @@
+require "config/rack/reloader"
+
+::Hanami.plugin do
+  env = ::Hanami.env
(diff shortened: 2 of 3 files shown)


pkuczynski:
This looks pretty cool, thank you! It looks that hanami 2.0.0.alpha1 is already using newer dry-validation, so this could work.

Could you please also add an example app to the tests, so we can test it against this newer hanami? You can have a look at how this has been done with rails using appraisal.

ippachi:
I'll do!

ippachi:
`2.0alpha` doesn't even seem to be implemented with the `new` command, so it's hard to add tests at this point.
So I will add a test when `2.0beta` comes out.

pkuczynski:
Sure thing. We could also add this to the earlier version of config gem (I think we added dry-validation in 2.1), but not sure if it's worth it. Better wait I guess...

Thanks for following up!

pkuczynski:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 15
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/piotr-kuczynski/config/pullRequest?prid=5743893&bid=18791066)

PR #274, opened 2020-03-31
Update default settings files for README

soartec-lab:
Thanks for making great gem.
While I am setting according to the README
I created this PR because I was worried.

## Summary
Update README.
Add 'config/settings.local.yml' into default settings files.

## Operation
```shell
$ rails new
# gem 'config' into Gemfile
$ bundle install
$  rails g config:install
      create  config/initializers/config.rb
      create  config/settings.yml
      create  config/settings.local.yml
      create  config/settings
      create  config/settings/development.yml
      create  config/settings/production.yml
      create  config/settings/test.yml
      append  .gitignore
```

## Enviroments
```shell
$ ruby -v => ruby 2.6.3p62 (2019-04-16 revision 67580)
$ rails -v => Rails 6.0.0
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Add `gem 'config'` to your `Gemfile` and run `bundle install` to install it. The
 which will generate customizable config file `config/initializers/config.rb` and set of default settings files:
 
     config/settings.yml
+    config/settings.local.yml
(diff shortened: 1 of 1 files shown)


pkuczynski:
Thanks

PR #277, opened 2020-06-02
Added alias to_h for to_hash

### LINKED ISSUE #217: Config::Options: to_hash and to_h have different behaviour
I would expect `to_h` and `to_hash` to have the same behaviour, however looks like this gem only overrides the definition of `to_hash` and does not alias that definition with `to_h`.

```
>> Settings.elasticsearch.to_h
=> {:host=>"localhost", :log=>true, :test=>#<Config::Options test=true>}
>> Settings.elasticsearch.to_hash
=> {:host=>"localhost", :log=>true, :test=>{:test=>true}}
```

Happy to file a patch with the alias - just wanted to check this wasn't the desired behaviour.

StoneGod:
Fixes #217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug fixes
+
+* Added alias to_h for to_hash ([#277](https://github.com/railsconfig/config/issues/277))
+
 ### Changes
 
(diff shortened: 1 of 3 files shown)


Fryguy:
@pkuczynski This PR seems to have broken backward compatibility - more details here: https://github.com/rubyconfig/config/issues/217#issuecomment-741953382

PR #279, opened 2020-06-09
Removes emojis from the post-install message

eksoverzero:
# Summary

Simpler terminals, like on servers or Docker build systems, crash due to not supporting emojis. It may be easier to just remove them rather than having to add support to these terminals for a single gem.


pkuczynski:
Fair enough!

PR #285, opened 2020-11-06
Rails alternative environment and additional key lookup during env mapping

neilwilliams:
There's three main changes in this PR.

1. A rails app can now configure where the 'environment' comes from. Previously, it was inferred from Rails.environment. We didn't want to pollute RAILS_ENV with actual environments (e.g. stage, preprod). So this change allows the two things to be separated. 

2. Devs are now able to run ```bundle exec rspec``` when using the ruby platform

3. When overriding keys from the environment, there wasn't a way to do this if the keys used in the settings.yml contained the seperation character being used in the environment variables.

E.g.
settings.yml
```
my_config:
  my_second_level_config:
    my_value: "1"
```

When using:
```
config.env_separator = '_'
```
I was never able to overide these settings with: 
```
ENV['MY_CONFIG_MY_SECOND_LEVEL_CONFIG_MY_VALUE'] = "2"
```

The change allows this to work as expected, but it will still allow the gem to work as it did before. 
E.g.
If the structure above cannot be found, it will set the value under:
```
my:
  config:
    my:
      second:
        level:
          config:
            my:
              value: ""
```

**Note** The code I've created for the third feature is quite crude at the moment, so would appreciate any feedback you have!


config.gemspec
```diff
@@ -37,9 +37,15 @@ Donate: \e[34mhttps://opencollective.com/rubyconfig/donate\e[0m\n"
...
   unless ENV['APPRAISAL_INITIALIZED'] || ENV['GITHUB_ACTIONS']
-    gems_to_install = /gem "(.*?)", "(.*?)"(?!, platform: (?!\[:ruby\]))/
-    File.read(Dir['gemfiles/rails*.gemfile'].sort.last).scan(gems_to_install) do |name, version|
-      s.add_development_dependency name, version
+    gems_to_install = /gem "(.*?)", "(.*?)"(, platform: \:(.*))?/
+    File.read(Dir['gemfiles/rails*.gemfile'].sort.last).scan(gems_to_install) do |name, version, _, platform|
+      if platform.nil?
+        s.add_development_dependency name, version
+      else
+        if platform.to_s == RUBY_ENGINE
```
Fryguy, line 45: Minor, but prefer `elsif` here.

config.gemspec
```diff
@@ -37,9 +37,15 @@ Donate: \e[34mhttps://opencollective.com/rubyconfig/donate\e[0m\n"
 
   # Default RSpec run will test against latest Rails app
   unless ENV['APPRAISAL_INITIALIZED'] || ENV['GITHUB_ACTIONS']
-    gems_to_install = /gem "(.*?)", "(.*?)"(?!, platform: (?!\[:ruby\]))/
-    File.read(Dir['gemfiles/rails*.gemfile'].sort.last).scan(gems_to_install) do |name, version|
-      s.add_development_dependency name, version
+    gems_to_install = /gem "(.*?)", "(.*?)"(, platform: \:(.*))?/
+    File.read(Dir['gemfiles/rails*.gemfile'].sort.last).scan(gems_to_install) do |name, version, _, platform|
```
Fryguy, line 41:
```suggestion
    gems_to_install = /gem "(.*?)", "(.*?)"(?:, platform: \:(.*))?/
    File.read(Dir['gemfiles/rails*.gemfile'].sort.last).scan(gems_to_install) do |name, version, platform|
```

I think you can do this to avoid capturing that third unused value

Fryguy:
I only commented on the Ruby env fix...can you please move that into a separate PR, as that is more likely to be merged sooner, and it's unrelated to this PR.

PR #289, opened 2020-11-25
Add exit! to reserved keywords

cjlarose:
Calling `Config::Options#[]` with the parameter `'exit!'` will invoke `Kernel#exit!` and immediately terminate the program. Instead, we want to handle this key explicitly, not because it's likely to be used as a key in a legitimate config yaml file, but to prevent a potential DOS vector if accessing options with untrusted input.


pkuczynski:
Makes sense! You forgot to add CHANGELOG entry. Just fixed it in your branch :)

pkuczynski (approved):

PR #292, opened 2020-12-01
Add explicit implementation of Config::Options#as_json

### LINKED ISSUE #287: Unexpected JSON serialisation behaviour
I ran into a weird issue when upgrading a Rails service recently.

Long story short, jbuilder has changed the way a template is transformed to JSON over the last couple of versions. See https://github.com/rails/jbuilder/blob/master/lib/jbuilder.rb#L250-L252 and the history on that line for details.

```ruby
# jbuilder v2.8.0
JSON.load Jbuilder.new { |json| json.config Config::Options.new(foo: :bar) }.target!
# => {"config"=>{"foo"=>"bar"}}

# jbuilder v2.9.1
JSON.load Jbuilder.new { |json| json.config Config::Options.new(foo: :bar) }.target!
# => {"config"=>[["foo", "bar"]]}
```

That was kind of unexpected. The reason for that is, that `Config::Options` includes `Enumerable` and [`Enumerable#as_json`](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/object/json.rb#L133-L137) calls `to_a` before serialising into JSON.

My current workaround would be to call `to_h`/`to_hash` on the `Config::Options` that's rendered or maybe even monkey-patching `Config::Options#as_json`.

I think it would be a good idea to implement `Config::Options#as_json`, maybe even removing `Config::Options#to_json`? Though that's a breaking change and would require a major version release? What do you all think?

cjlarose:
Fixes #287 

Before, `Config::Options#as_json` was not explicitly defined, but the class included the `Enumerable` module. So, if you were using ActiveSupport Core Extensions, then the definition of `#as_json` for Enumerable was invoked, which would represent the `Config::Options` as a Array of key-value pairs instead of a Hash.

This changes brings `#as_json` in line with the output we get from `to_json`: essentially, calling these methods on an instance of `Config::Options` behaves the same as if they were called instead on the hash returned by `to_hash`.

It's not clear if this is a "bug fix" or a "breaking change" from a semver perspective, but I think it's unlikely that someone was depending on the output from `#as_json` to be the Array-of-pairs representation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Added alias to_h for to_hash ([#277](https://github.com/railsconfig/config/issues/277))
 * Prevent unnecessary doubled loading of environment variables ([#291](https://github.com/rubyconfig/config/pull/291))
+* Return `Hash` from `Config::Options#as_json` instead of `Array` of pairs when using ActiveSupport Core Extensions ([#292](https://github.com/rubyconfig/config/pull/292))
 
 ### Changes
 
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -129,6 +129,10 @@ module Config
       to_hash.to_json(*args)
     end
 
+    def as_json(options = nil)
+      to_hash.as_json(options)
+    end
+
     def merge!(hash)
       current = to_hash
       DeepMerge.deep_merge!(
--- a/spec/options_spec.rb
+++ b/spec/options_spec.rb
@@ -224,4 +224,10 @@ describe Config::Options do
 
   end
 
+  context 'when calling #as_json' do
+    it 'should return a Hash of the keys and values' do
+      options = Config::Options.new(foo: :bar)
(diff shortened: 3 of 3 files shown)


cjlarose:
Tests fail when using the Sinatra Gemfiles, which is more-or-less expected: if you don't have the ActiveSupport Core Extensions loaded, then `Hash#as_json` isn't defined.

I don't know yet what the right fix is. I'm considering

1. Skip the test except when testing against Rails
2. Detect at runtime whether or not the Hash returned by `to_hash` responds to `as_json` and do something else if it doesn't . Maybe just return `to_hash` itself?

cjlarose:
@pkuczynski Updated to skip the test except when running against Rails. LMK what you think!

pkuczynski:
Looks very good to me! I will merge it in and release new version :) Thanks for all your hard work on this @cjlarose!

pkuczynski (approved):

PR #293, opened 2020-12-09
Raise explicit error on env var conflicts

### LINKED ISSUE #205: no implicit conversion of Symbol into Integer when config.use_env = true
Hello, 

I found this problem with a specific combination of ENV variables like: 

```
BACKEND_DATABASE=development
BACKEND_DATABASE_USER=postgres
```

When iterating on the second env variable [lib/config/options.rb](https://github.com/railsconfig/config/blob/master/lib/config/options.rb#L62) raises  `no implicit conversion of Symbol into Integer` when trying to find `leaf[keys.last]` but leaf is a string containing `development`. 

My workaround for this is to make sure the ENV variables don't get into this form of nesting. 

```
BACKEND_DATABASE_NAME=development
BACKEND_DATABASE_USER=postgres
```

It's probably reasonable to require this form in the docs, as I wouldn't imagine how to create a structure where one of the keys contains both data and other keys.

cjlarose:
Fixes #205

Suppose the user sets the variables `Settings.backend_database` in addition to `Settings.backend_database.user`. Since it is possible to override existing settings with env vars, one possible way to handle this is to process the environment variables in a well-defined order and allow lexicographically greater keys to override lower ones. However, it's probably not what the user intended to happen, so instead we raise a descriptive error when a conflict is detected.

This is likely not a breaking change because a `RuntimeError` is already raised in this case, it's just not very friendly:

    TypeError:
      no implicit conversion of Symbol into Integer

Now we raise a nicer error

    RuntimeError:
        Environment variable Settings.backend_database.user conflicts
        with variable Settings.backend_database

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 ## Unreleased
 
+### Bug fixes
+
 ...
+
+### Changes
+
+* Raise explicit error on environment variable conflicts ([#293](https://github.com/railsconfig/config/issues/293))
 
 ## 2.2.2
 
--- a/README.md
+++ b/README.md
@@ -420,6 +420,21 @@ ENV['Settings.section.server'] = 'google.com'
 
 It won't work with arrays, though.
 
+It is considered an error to use environment variables to simutaneously assign a "flat" value and a multi-level value to a key.
+
+```ruby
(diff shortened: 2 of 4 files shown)


pkuczynski (approved):
Fantastic work! Thanks :)

Issue #271, opened 2020-01-28
AWS secrets manager?


hopewise:
Currently we use Capistrano to deploy our app to AWS, which we can handle to symlink `config/settings/production.yml`, however, we are studying the possibility of using AWS [Fargate](https://aws.amazon.com/fargate/),  and it seems that AWS [Secrets Manager](https://aws.amazon.com/secrets-manager/) is recommended when using Fargate.

Not to mention that maximum secret size for AWS Secrets Manager in 10kb, while our `production.yml` size is about 50kb, I will have to re-fill it in multiple secrets .. there must be work around for this..

The question is, what's the possibilities to use AWS Secrets Manager to populate the settings behind the scenes rather than relying on files at for ex: `config/settings/production.yml` ? so that we keep our code not changed, ex: `Settings.stripe.publishable_key` ?

pkuczynski:
I think the easiest option would be to expose secrets from AWS Secrets Manager to Config using environment variables. Have a look at https://github.com/rubyconfig/config#working-with-environment-variables to see how to automatically populate ENV into settings

pkuczynski:
And if you find something which is not mentioned in the documentation, please share it with the community via pull request to README :)

hopewise:
Thanks, considering aws pricing $0.4 per a secret, its too expensive comparing to the number of secrets we have! I must find a solution, like grouping each 100 secret or so in one secret.. or completely drop the aws secrets manager idea..

cjlarose:
A fix for this was implemented in #299 and released in config version 3.0.0.

PR #299, opened 2021-01-13
Refactor environment variable handling into new Sources::EnvSource

cjlarose:
In principle, loading settings from environment variables isn't much different from loading them from a `YAMLSource` or a `HashSource`. This change unifies the handling of environment variables into the existing "sources" abstraction.

My own personal motivation for this change is essentially the same as https://github.com/rubyconfig/config/issues/271. I want to load settings into `config` that I got from AWS Secrets Manager. Although AWS Secrets Manager's API _technically_ allows you to store arbitrary string values, the AWS Console UI and the AWS Secrets Manager API encourages you to store your application's secret configuration as as single Secret resource that is a flat, JSON-encoded key-value mapping. This maps very nicely to how one might use environment variables to configure an application.

### Editing secrets in the AWS console

Screen Shot 2021-01-12 at 20 04 41

### Actual plaintext

Screen Shot 2021-01-12 at 20 10 37

The problem is that while `config` offers the ability to parse nested settings from `ENV`, it doesn't allow you to do the same for settings you got from another source, like AWS Secrets Manager. As a result, users are forced into an awkward workaround: export all secrets from AWS Secrets Manager into `ENV`, then ask `config` to load the `ENV` again. My proposed `EnvSource` allows users to skip the loading-into-ENV step and just hand the flat `Hash` over to `config` directly.

 For example,
 
 ```ruby
# fetch secrets from AWS
client = Aws::SecretsManager::Client.new
response = client.get_secret_value(secret_id: "#{ENV['ENVIRONMENT']}/my_application")
secrets = JSON.parse(response.secret_string)

# load secrets into config
secret_source = Config::Sources::EnvSource.new(secrets)
Settings.add_source!(secret_source)
Settings.reload!
```

### DIFF
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -31,47 +31,6 @@ module Config
       @config_sources.unshift(source)
     end
 
-    def reload_env!
-      return self if ENV.nil? || ENV.empty?
-
-      hash = Hash.new
-
-      ENV.each do |variable, value|
-        separator = Config.env_separator
-        prefix = (Config.env_prefix || Config.const_name).to_s.split(separator)
-
-        keys = variable.to_s.split(separator)
-
-        next if keys.shift(prefix.size) != prefix
-
-        keys.map! { |key|
-          case Config.env_converter
-            when :downcase then
-              key.downcase.to_sym
-            when nil then
-              key.to_sym
-            else
-              raise "Invalid ENV variables name converter: #{Config.env_converter}"
-          end
-        }
-
-        leaf = keys[0...-1].inject(hash) { |h, key|
-          h[key] ||= {}
-        }
-
-        unless leaf.is_a?(Hash)
-          conflicting_key = (prefix + keys[0...-1]).join(separator)
-          raise "Environment variable #{variable} conflicts with variable #{conflicting_key}"
-        end
-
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: lib/config/options.rb [hunk 1]
```diff
@@ -31,47 +31,6 @@ def prepend_source!(source)
...
-          raise "Environment variable #{variable} conflicts with variable #{conflicting_key}"
-        end
-
-        leaf[keys.last] = Config.env_parse_values ? __value(value) : value
-      end
-
-      merge!(hash)
-    end
-
-    alias :load_env! :reload_env!
```
cjlarose, line 73: @pkuczynski `Config::Options#load_env!` and `Config::Options#reload_env!` are not documented at all, but are technically `public` methods.

As such, depending on whether or not we consider all visible methods to be part of the public API of this gem, this could be considered a breaking change and warrants a major version bump.

The change for users that use `#load_env!` or `#reload_env!` is simple: they should just replace those usages with `#reload!`, which is documented.

cjlarose:
@pkuczynski let me know if you have any concerns

cjlarose:
Ended up speaking to @pkuczynski about this in another channel. He's cool with merging and releasing.

This change was released in 3.0.0.

Issue #288, opened 2020-11-24
Optional namespace for `add_source!`


wwboynton:
It would be nice to be able to invoke `Settings.add_source!` with an optional parent prefix such that new sources could be loaded in programmatically into different levels of the Settings structure. This would allow for multiple YML files of similar function to be grouped under the same parent without having to specify that namespace in every single YAML file.

In my case, I have a lot of user-defined yml files together in a directory which are split up for organization/legibility, but which cohabitate in the same parent node of the `Settings` object when all loaded together. It's confusing to ask my users to specify a few levels of parent nodes at the top of every single file in the directory just so that the `Settings` object will be correctly organized when it gets to my code. It would be nice to write code which will load these YAML files into the `Settings` object in a way that mimics their directory structure on the filesystem.

Modifying the `yaml_source.rb` source to accommodate this might look something like this:

```
# config/lib/config/sources/yaml_source.rb 

module Config
  module Sources
    class YAMLSource
      attr_accessor :path

      def initialize(path, namespace = nil)  # New variable here
        @path = path.to_s
        @namespace = namespace 
      end

      # returns a config hash from the YML file
      def load
        result = YAML.load(ERB.new(IO.read(@path)).result) if @path and File.exist?(@path)

        if !@namespace.nil? && result        # New logic here
          result = @namespace.reverse.inject(result) { |a, n| { n => a } }
        end
        result || {}
        rescue Psych::SyntaxError => e
      
        # ...etc

```

Running that like so:

```
puts "Without Namespace!"
y = Config::Sources::YAMLSource.new("/Users/userguy/repos/scratch/example.yml")
pp y.load

puts "With Namespace!"
y = Config::Sources::YAMLSource.new("/Users/userguy/repos/scratch/example.yml", ['new-level'])
pp y.load
```

...produces output like this:
```
Without Namespace!
{"foo"=>["bar", "baz", {"chungus"=>["Bungus"]}]}
With Namespace!
{"new-level"=>{"foo"=>["bar", "baz", {"chungus"=>["Bungus"]}]}}
```

If you'd like me to take a crack at doing this and submitting a PR, let me know and I'm happy to do it. Maybe I am just very tired and completely missed an existing way to do this in the documentation, or am overthinking this and forgetting something fundamental about YAML that will solve my problem. It has been known to happen, in which case please close this issue and accept my apologies for wasting your time!

pkuczynski:
[...]

PR #301, opened 2021-03-06
Add namespaces to YAML sources per #288

wwboynton:
This PR adds functionality for optional namespaces when creating a new `Config::Sources::YAMLSource` source or using `Settings.add_source!` per discussion in #288.

The functionality works like this:

Given a YAML file like this:
```yaml
foo:
  - bar
  - baz
  - chungus:
    - Bungus
```

The code works like this:
```ruby
# No namespace, same as current functionality
y = Config::Sources::YAMLSource.new("/tmp/example.yml")
y.load
=> {"foo"=>["bar", "baz", {"chungus"=>["Bungus"]}]}

# Single namespace, array or string both fine
y = Config::Sources::YAMLSource.new("/tmp/example.yml", 'new-level')
y.load
=> {"new-level"=>{"foo"=>["bar", "baz", {"chungus"=>["Bungus"]}]}}

# Nested namespace, given as array of keys
y = Config::Sources::YAMLSource.new("/tmp/example.yml", ['new-level', 'level-2'])
y.load
=> {"new-level"=>"level-2"=>{"foo"=>["bar", "baz", {"chungus"=>["Bungus"]}]}}}
```

Tests have been written and run to cover both cases based entirely on the existing `basic yml file` test case, called `basic yml file with single namespace` and `basic yml file with nested namespace` respectively.

I'm happy to make any changes necessary for code style (i.e. keyword method argument, ternary as opposed to multi-line, whatever) or functionality in keeping with existing standards for this codebase of which I may be ignorant.

### DIFF
--- a/lib/config/sources/yaml_source.rb
+++ b/lib/config/sources/yaml_source.rb
@@ -6,15 +6,26 @@ module Config
     class YAMLSource
       attr_accessor :path
 
-      def initialize(path)
+      def initialize(path, namespace = nil)
         @path = path.to_s
+
+        # Make sure @namespace is an array if it exists at all
+        if namespace
+          @namespace = namespace
+          @namespace = [@namespace] unless @namespace.is_a?(Array)
+        end
       end
 
       # returns a config hash from the YML file
       def load
         result = YAML.load(ERB.new(IO.read(@path)).result) if @path and File.exist?(@path)
-
-        result || {}
+        return {} unless result
+        return result unless @namespace
+        
+        # Resolves namespacing multiple layers deep 
+        # i.e. ['layer1', 'layer2'] comes out to {'layer1' => {'layer2' => content}}
+        return  result = @namespace.reverse.inject(result) { |a, n| { n => a } }
+        
 
         rescue Psych::SyntaxError => e
           raise "YAML syntax error occurred while parsing #{@path}. " \
--- a/spec/sources/yaml_source_spec.rb
+++ b/spec/sources/yaml_source_spec.rb
@@ -25,6 +25,43 @@ module Config::Sources
       end
     end
 
+    context "basic yml file with single namespace" do
+      let(:source) do
+        YAMLSource.new "#{fixture_path}/development.yml", 'test_namespace'
+      end
+
+      it "should properly read the settings" do
+        results = source.load
+        expect(results['test_namespace']["size"]).to eq(2)
+      end
+
+      it "should properly read nested settings" do
+        results = source.load
+        pp results
+        expect(results['test_namespace']["section"]["size"]).to eq(3)
+        expect(results['test_namespace']["section"]["servers"]).to be_instance_of(Array)
+        expect(results['test_namespace']["section"]["servers"].size).to eq(2)
+      end
+    end
+
+    context "basic yml file with nested namespace" do
+      let(:source) do
+        YAMLSource.new "#{fixture_path}/development.yml", ['test_namespace', 'test_layer_2']
+      end
+
+      it "should properly read the settings" do
+        results = source.load
+        expect(results['test_namespace']['test_layer_2']["size"]).to eq(2)
+      end
+
+      it "should properly read nested settings" do
+        results = source.load
+        expect(results['test_namespace']['test_layer_2']["section"]["size"]).to eq(3)
+        expect(results['test_namespace']['test_layer_2']["section"]["servers"]).to be_instance_of(Array)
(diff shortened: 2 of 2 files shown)


wwboynton:
Looks like it's failing in 2.4...I guess I'll look at that.

wwboynton:
Seems like there may be issues with 2.4 writ-large. I thought maybe I'd used some new-fangled syntax somewhere so I opted to just install 2.4.0 and run the tests. Short version, I can't `bundle install` under 2.4.0 because of a dependency on zeitwerk:
```bash
Resolving dependencies...
zeitwerk-2.4.2 requires ruby version >= 2.4.4, which is incompatible with the
current version, ruby 2.4.0p0
```

Moving to 2.4.4 produces a dependency in `activesupport-6.0.3.1` on ruby 2.5:
```bash
wwboynton@camelopardalis:config$ bi
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
activesupport-6.0.3.1 requires ruby version >= 2.5.0, which is incompatible
with the current version, ruby 2.4.4p296
```

So maybe it's time to sunset support for 2.4 in this gem after all. That or downgrade some stuff until it works again.

wwboynton:
Deleting previous comments -- I'm an idiot, and the problem is my fault. In trying to figure out why 2.4 tests were failing, I didn't wipe my Gemfile.lock between ruby versions, causing me to stay locked on new and incompatible gem versions with an older ruby. I resolved the issue by being less of an idiot and looking at the logs, discovering that the root cause in the first place was that I'd left a `pp` line I had used for debugging in the spec, and 2.4 was before the time that `pp` didn't need to be required explicitly. So, in a way, running it through an old version of ruby helped catch my dumb embarrassing mistake.

Anyway, the tests appear to pass now in native ruby 2.4-2.7, so sorry to anyone who had to read my drivel in this thread earlier. Thanks again for your encouragement and I hope this code is useful now that it's received some more scrutiny and caffeine.

wwboynton:
I guess I can't see why one wouldn't want the same functionality to be consistent for hash sources since they get used similarly, so for consistency I went ahead and replicated the work in the same way for the hash sources and tests.

pkuczynski:
@cjlarose will you have some time to look into this one?

cjlarose:
> @cjlarose will you have some time to look into this one?

Yep! I can take a look this week.

wwboynton:
> > @cjlarose will you have some time to look into this one?
> 
> Yep! I can take a look this week.

Awesome, thank you! Let me know if there's anything I should do differently and I'm happy to accommodate as time allows.

cjlarose:
Hey @wwboynton sorry it took so long to get back to you! I ended up having a busy week.

Anyway, I love the enthusiasm and I like the overall idea. However, I had an idea that I wanted to run by you because I think it'll clean up the code a little and also make this feature available more broadly. For example, there's a new `Sources::EnvSource` that I added in #299 and users could have their own custom `Source`s, but I think the idea of prefixing stuff with a namespace could be useful for all kinds of `Source`s. Instead of modifying the existing `Source` implementations, I think we can use some good ol' fashioned composition here.

Remember that a `source` from the perspective of the `config` gem is just something that responds to `.load` and returns a `Hash`, whenever the user first initializes their settings or calls `Settings.reload!`. Using that, I think it's possible to write a new, higher-order `source` class that takes a namespace _and another source_ to add the namespace to. So you could do something like

```ruby
original_source = Config::Sources::YAMLSource.new "/path/to/source.yml"
namespaced_source = Config::Sources::NamespacedSource.new ['new-level', 'level-2'], original_source
Settings.add_source! namespaced_source
```

We could still offer the shorthand version with the optional namespace parameter `Settings.add_source! "/path/to/source.yml", ['new-level', 'level-2']`, but under the hood, it'd use the new `NamespacedSource`.

Of course, the new `NamespacedSource` just has to implement a `load` method itself, which would just invoke `load` on the original source and then add the namespace on top of it before returning it.

Let me know if that makes sense, and more importantly, if it'd address your original use case.

wwboynton:
@cjlarose hilariously, I don't think I ever saw this reply on this PR. Sorry about that! I came back today to report a small bug I found and stumbled upon this by accident, lol.

I don't have time to build that solution out, but I can definitely confirm that it would have solved my original use-case and I think it definitely makes the library more flexible for handling multiple YAML files without a bunch of unnecessary levels of keys in the files themselves.

PR #303, opened 2021-04-09
Allow config option to disable ERB in YAML

kmcphillips:
## Problem

There is currently no way to disable ERB evaluation in YAML files at load time.

This is an issue if the config file has a string that contains ERB to be evaluated later, or something that looks like ERB. It simply raises an error (such as `SyntaxError` or `RuntimeError`) at application load time.

## Solution

Add the `Config.evaluate_erb_in_yaml` option, and use it to conditionally evaluate ERB or simply pass the string through.

This is fully backwards compatible as the value defaults to `true`.

### DIFF
--- a/lib/config/sources/yaml_source.rb
+++ b/lib/config/sources/yaml_source.rb
@@ -12,7 +12,11 @@ module Config
 
       # returns a config hash from the YML file
       def load
-        result = YAML.load(ERB.new(IO.read(@path)).result) if @path and File.exist?(@path)
+        if @path and File.exist?(@path)
+          file_contents = IO.read(@path)
+          file_contents = ERB.new(file_contents).result if Config.evaluate_erb_in_yaml
+          result = YAML.load(file_contents)
+        end
 
         result || {}
 
--- a/lib/config/sources/yaml_source.rb
+++ b/lib/config/sources/yaml_source.rb
@@ -5,16 +5,18 @@ module Config
   module Sources
     class YAMLSource
       attr_accessor :path
+      attr_reader :evaluate_erb
 
-      def initialize(path)
+      def initialize(path, evaluate_erb: Config.evaluate_erb_in_yaml)
         @path = path.to_s
+        @evaluate_erb = !!evaluate_erb
       end
 
       # returns a config hash from the YML file
       def load
         if @path and File.exist?(@path)
           file_contents = IO.read(@path)
-          file_contents = ERB.new(file_contents).result if Config.evaluate_erb_in_yaml
+          file_contents = ERB.new(file_contents).result if evaluate_erb
           result = YAML.load(file_contents)
         end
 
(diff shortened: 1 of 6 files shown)


lib/config/sources/yaml_source.rb
```diff
@@ -12,7 +12,11 @@ def initialize(path)
 
       # returns a config hash from the YML file
       def load
-        result = YAML.load(ERB.new(IO.read(@path)).result) if @path and File.exist?(@path)
+        if @path and File.exist?(@path)
+          file_contents = IO.read(@path)
+          file_contents = ERB.new(file_contents).result if Config.evaluate_erb_in_yaml
```
cjlarose, line 17: Minor suggestion that would be really nice: give users the option to opt into- or out of this behavior on a per-file basis if they're using the `YAMLSource` API directly. For example,

```ruby
Config.add_source!(Config::Sources::YAMLSource.new(path1)) # defaults to global Config.evaluate_erb_in_yaml
Config.add_source!(Config::Sources::YAMLSource.new(path1, evaluate_erb: false)) # overrides global setting
Config.add_source!(Config::Sources::YAMLSource.new(path1, evaluate_erb: true)) # overrides global setting
```

That way it's not all-or-nothing. We do something similar for `EnvSource` here: configuration is derived from the global values if left unspecified, but can always be overridden. 

https://github.com/rubyconfig/config/blob/618c839ccca454d419b0587f8a113e2fe6f3ebcd/lib/config/sources/env_source.rb#L10-L14

cjlarose (commented):
Thanks for your contribution! This is great!

I added a suggestion to make it a little nicer if you're up to it. Otherwise, I'm happy to merge as-is and leave the suggestion for future work.

lib/config/sources/yaml_source.rb
kmcphillips: Oh neat. Ok I'll spend some time with this. Feedback makes sense.

kmcphillips:
I went back and forth on the config name. `evaluate_erb` is simpler, but then I realized that it only applies to YAML source, so I set it to `evaluate_erb_in_yaml`. Thoughts?

kmcphillips:
Or I guess a third option is `evaluate_erb` in the `YAMLSource` class, and `evaluate_erb_in_yaml` in the global config.

kmcphillips:
Ready for another review.

cjlarose:
> Or I guess a third option is `evaluate_erb` in the `YAMLSource` class, and `evaluate_erb_in_yaml` in the global config.

I like this best and it looks like this is what you went with.

Code looks great! Thanks for making the changes. I'll do some local testing here in the next couple of days, merge, and publish a new version.

kmcphillips:
Ok! Let me know if you want changes. I'm happy with it I think, but every codebase is different.

Thanks for accepting the contribution.

cjlarose (approved):
Awesome work!

kmcphillips:
Nice! Thanks!

cjlarose:
Published in version 3.1.0!

Issue #286, opened 2020-11-09
allow symbol value via ENV


khoan:
Setup

```
cd /path/to/rails/app/src
cat config/initializers/config.rb
Config.setup do |config|
  config.const_name = 'Settings'
  config.use_env = true
  config.env_prefix = 'SETTINGS'
  config.env_separator = '__'
  config.env_converter = :downcase
  config.env_parse_values = true
end
cat config/settings/test.yml
provider: :bam
```

```
SETTINGS__PROVIDER=:bam RAILS_ENV=test bin/rails c
Settings.provider # => ":bam"   want symbol instead of string
```

```
RAILS_ENV=test bin/rails c
Settings.provider # => :bam  got symbol as expected
```

maybe related to [issue#194](https://github.com/rubyconfig/config/issues/194)? Thanks.

pkuczynski:
I don't think its related to #194. They actually wanted something opposite to your case.

Env vars are string and I don't think we should force conversion to symbols. However we could implement a logic where strings staring from `:` would be converted to symbols (see: https://stackoverflow.com/questions/2004491/convert-string-to-symbol-able-in-ruby)

If you are interested, maybe you can fire up a PR?

pkuczynski:
@khoan are you interested in firing up PR for this?

manjunath724:
@pkuczynski 

I always get Symbol `:bam` for the above configuration. I am executing this on `rails-6.1.4` & `ruby-3.0.2`. Can you please help me replicate? I'd be more than happy to create a PR.

cjlarose:
@manjunath724 If I understand the feature request correctly, this can be replicated with the following spec:

```diff
diff --git a/spec/config_env_spec.rb b/spec/config_env_spec.rb
index 65e7da8..5f91def 100644
--- a/spec/config_env_spec.rb
+++ b/spec/config_env_spec.rb
@@ -77,6 +77,12 @@ describe Config::Options do
           expect(config.new_var.is_a? Float).to eq(true)
         end
 
+        it 'should recognize strings starting with : as symbols' do
+          ENV['Settings.new_var'] = ':hello'
+
+          expect(config.new_var).to eq(:hello)
+        end
+
         it 'should leave strings intact' do
           ENV['Settings.new_var'] = 'foobar'
 
```

Execute

```
bundle install
bundle exec appraisal install
bundle exec appraisal rails-6.1 rspec spec/config_env_spec.rb
```

And you'll get this result

```
Bundler version 2.1.4
[...]

PR #309, opened 2021-10-06
Allow symbol value via ENV

manjunath724:
This PR converts string starting with `:` to symbol. Allows symbol value via ENV.

`Config::Options` when overriding settings via ENV variables is enabled and parsing ENV variable values is enabled should recognize strings starting with `:` as symbols.

But strings starting with :: will be left intact.

Ex:
```
# Given the following environment:
SETTINGS__SECTION__SERVER_TYPE=:remote
SETTINGS__SECTION__SERVER_ACCESS=::ENABLED
```

```
# The following settings will be available:
Settings.section.server_type # => :remote
Settings.section.server_access # => '::ENABLED'
```

`String#strip` is used for removing whitespace characters from string inputs, such as: `'   :remote   '`, `'  :   '`, `'    :     remote   '`

`.gsub` is used to replace characters with the help of regex pattern match

`.tr` to covert to snakecase

`.downcase` to transform uppercase letters to lowercase

Ref: https://github.com/rubyconfig/config/issues/286

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,7 +461,8 @@ For instance, given the following environment:
 SETTINGS__SECTION__SERVER_SIZE=1
 SETTINGS__SECTION__SERVER=google.com
 SETTINGS__SECTION__SSL_ENABLED=false
-SETTINGS__SECTION__SERVER_ACCESS=:remote
+SETTINGS__SECTION__SERVER_TYPE=:remote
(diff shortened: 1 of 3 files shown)


cjlarose:
Closing because this is stale. I'm open to the idea, but the particular implementation here needs some work.

PR #310, opened 2021-10-10
DRY - Code Refactor

manjunath724:
- `case..when..else..end` blocks have been aligned vertically
- Moved common code snippets to a private methods
- Used `eval(str)` to return Boolean `true/false`
- Removed `!!`(Double not) prepended to `true/false`

### DIFF
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -15,20 +15,11 @@ module Config
     end
 
     def add_source!(source)
-      # handle yaml file paths
-      source = (Sources::YAMLSource.new(source)) if source.is_a?(String)
-      source = (Sources::HashSource.new(source)) if source.is_a?(Hash)
-
-      @config_sources ||= []
-      @config_sources << source
+      load_config_sources(:push, source)
     end
 
     def prepend_source!(source)
-      source = (Sources::YAMLSource.new(source)) if source.is_a?(String)
-      source = (Sources::HashSource.new(source)) if source.is_a?(Hash)
-
-      @config_sources ||= []
-      @config_sources.unshift(source)
+      load_config_sources(:unshift, source)
     end
 
     # look through all our sources and rebuild the configuration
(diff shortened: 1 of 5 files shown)


manjunath724:
While working PR https://github.com/rubyconfig/config/pull/309. I noticed that a lot of code could be refactored - which has led to this PR https://github.com/rubyconfig/config/pull/310. Can we please approve the tests to run on this branch?

lib/config/options.rb
```diff
@@ -152,23 +121,46 @@ def respond_to_missing?(*args)
...
+                            dest,
+                            preserve_unmergeables: false,
+                            knockout_prefix:       Config.knockout_prefix,
+                            overwrite_arrays:      Config.overwrite_arrays,
+                            merge_nil_values:      Config.merge_nil_values,
+                            merge_hash_arrays:     Config.merge_hash_arrays
+                          )
+    end
+
+    def instance_of(value)
```
Fryguy, line 145: `instance_of` is a strange name here, but I'm struggling coming up with a better name.  To me, `instance_of` implies you are getting another instance of something (perhaps a dup), but it's actually returning the converted value of the passed in value.  Perhaps

```suggestion
    def serialize_value(value)
```

Once that is changed, I was thinking you can also rename `descend_array` to `serialize_array` to match the pattern, but now that I look at it, there's no longer a reason to have a separate method, and it can probably be inlined directly into this method.

lib/config/sources/yaml_source.rb
```diff
@@ -9,7 +9,7 @@ class YAMLSource
 
       def initialize(path, evaluate_erb: Config.evaluate_erb_in_yaml)
         @path = path.to_s
-        @evaluate_erb = !!evaluate_erb
+        @evaluate_erb = evaluate_erb
```
Fryguy, line 12: Not sure about this one...it's kind of nice having a proper boolean, since it is accessible via the attr_reader.

cjlarose:
Closing because this is stale

Issue #276, opened 2020-05-16
Split validation via dry-rb into seperate gem


pfeiffer:
Hi!

Since version 2.1 `config` supports validation via `dry-rb`. Unfortunately this dependency results in a quite high memory usage.

Output using `derailed bundle:mem`:

```
 rails/all: 25.1523 MiB
    rails: 9.4883 MiB
       ...
 config: 7.1172 MiB
    config/validation/schema: 6.6563 MiB (Also required by: config/validation/schema)
      dry-schema: 6.6367 MiB
        dry/schema: 6.6211 MiB
          dry/schema/dsl: 6.5039 MiB
            dry/schema/config: 4.3203 MiB (Also required by: dry/schema/rule_applier)
              dry/schema/type_container: 3.8398 MiB (Also required by: dry/schema/processor)
                dry/types: 3.4727 MiB (Also required by: dry/schema/types)
                  dry/types/nominal: 1.8398 MiB
                    dry/types/builder: 1.0703 MiB
                      dry/types/constrained: 0.707 MiB
                        dry/types/constraints: 0.6523 MiB
                          dry/logic/rule_compiler: 0.5977 MiB (Also required by: dry/schema/compiler)
                            dry/logic/rule: 0.5234 MiB (Also required by: dry/logic/rule/predicate)
                              dry/logic/operations: 0.4023 MiB
                    dry/types/array: 0.3008 MiB
                  dry/types/core: 0.6758 MiB
                  dry/types/inflector: 0.3984 MiB
                    dry/inflector: 0.3242 MiB
                dry/container: 0.3555 MiB (Also required by: dry-container, dry/types/container)
            dry/schema/macros: 1.0977 MiB
              dry/schema/macros/optional: 0.6094 MiB
                dry/schema/macros/key: 0.582 MiB (Also required by: dry/schema/macros/required)
                  dry/schema/processor: 0.4922 MiB (Also required by: dry/schema/dsl, dry/schema/params)
                    dry/schema/rule_applier: 0.4531 MiB (Also required by: dry/schema/dsl)
                      dry/schema/message_compiler: 0.3125 MiB
```

In our app with quite a lot of dependencies, this makes `config` the second most memory hungry gem after `rails`.

Would it be an idea to split the validation feature into a separate optional gem, that could be included if this functionality is wanted? This would cut down significantly on the memory impact of including the core `config` gem making inclusion of `dry-rb` optional for the developer.

pfeiffer:
[...]

PR #317, opened 2022-04-26
Remove `dry-validation` from dependencies

BuonOmo:
Fixes https://github.com/rubyconfig/config/issues/276

# Problem

Having the dependency on a fix dry-validation version generates multiple issues:

- users who want another version of the dry-validation gem are blocked
- the dry-validation gem itself has dependencies, making the package size bigger, even if not needed.

# Solution

We remove the dependency, but still check *at runtime* that the dependency is met *if and only if* the schema validation feature is used.

# Other aspects

All of the changes described below are in a separate commit, I can revert it if it helps merging!

## Ruby 2.1 compatibility

During the implementation, I had a bug with `Config` reloading. The `lib/config/compatibility.rb` introduces a bug that is really hard to debug (for me at least):

```ruby
# in config.gemspec
require_relative "config/version" # => true

# somewhere after
p Config.const_defined?(:CONFIG) # => true
require "config" # => true
# will remove Config and then require everything... Except what was 
# already required, e.g. "config/version"
p Config.const_defined?(:CONFIG) # => false
```

Since this file was added for an EOL ruby version (2.1, per commit https://github.com/rubyconfig/config/commit/c4119fb2a31f40759718a0169e2b59a6f86c9aad), I removed it.

## Usage of `require_relative`

Helps be sure that we talk about the same file! Otherwise, it depends on `$LOAD_PATH` (or `$:`)

## `Dir[..].sort.each { require }`

Make sure that the require order is not OS, or filesystem dependent

## `.ruby-version`

I've removed the file, as for `Gemfile.lock`, the ruby version will in the end depend on the person using the gem, so developers of the gem should also be able to run it with their local ruby versions!

## `Config::Error`

Since there are now two possible errors, lets regroup those under the same banner! (not that it is not a breaking change since `.is_a?` will still behave the same)


teohm:
What's the next step for this PR? We are waiting for the release of this PR, so that we can use RubyConfig in our production apps. Anything we can help to move the PR to the next stage?

Some context: Our production apps are using an older version of `dry-validation` gem.

BuonOmo:
@teohm same here, I think the most important thing needed is traction and a maintainer discussing it. If you think it needs something from my end as well, please ping me. 

I guess you could also use this branch, which would prove that it works fine :)

BuonOmo:
@pkuczynski @cjlarose sorry for the noise, but should I update the PR, do you think you'd be able to review it any time?

cjlarose:
@BuonOmo Sorry I dropped the ball on this. This looks great!

What is the experience if someone specifies a `validation_contract` instead of using `config.schema` as described in the README? Is a runtime check still performed to ensure we're using a compatible version?

BuonOmo:
Closed in favor of #333

PR #318, opened 2022-05-25
fix: remove deprecated method 'exists?' removed in ruby 3.0

danilogco:
It alllows to use the gem in a project based on Ruby > 2.7. The changed method doesn't exist anymore.

### DIFF
--- a/lib/generators/config/install_generator.rb
+++ b/lib/generators/config/install_generator.rb
@@ -18,7 +18,7 @@ module Config
       end
 
       def modify_gitignore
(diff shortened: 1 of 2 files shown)


rdubya (approved):

pkuczynski:
How would that affect earlier versions of ruby?

danilogco:
> How would that affect earlier versions of ruby?

The method ".exists?" is deprecated since the ruby 2.2.0 and it was removed in the most recent versions.

cjlarose (approved):

PR #335, opened 2023-04-04
Update and rename lib/config/validation/validate.rb to liberación/con…

fxckyuo:
…figuración/validación/validate.rb


cjlarose:
Not interested in this change

Issue #340, opened 2023-08-16
Allow to use custom filename && directory name to store configs


feruzoripov:
**Problem**

The current solution of hardcoding the filename and directory name to `settings.yml` and `settings` folder is not ideal, as it can lead to conflicts if the same filename or directory name is already being used by other logics.

A better solution is to add two configuration fields, `Config.file_name` and `Config.dir_name`, which will allow users to specify their own filename and directory name. This will make it possible to avoid conflicts and it will also give users more flexibility in how they store their data.

**Solution**

Added two configuration fields:

- `Config.file_name` for filename
- `Config.dir_name` for directory name

feruzoripov:
https://github.com/rubyconfig/config/pull/341

PR #341, opened 2023-08-16
Allow to use custom filename && directory name to store configs

feruzoripov:
https://github.com/rubyconfig/config/issues/340


lib/config/version.rb
```diff
@@ -1,3 +1,3 @@
 module Config
-  VERSION = '4.2.1'.freeze
```
pkuczynski, line 2: I would leave this to the maintainers, whenever they would be releasing the gem... Please revert...

pkuczynski (commented):
Can you also update the docs please (README)?

lib/config/version.rb
feruzoripov: reverted version and updated README @pkuczynski

cjlarose (approved):

cjlarose:
Released with version 5.2.0

PR #342, opened 2023-10-02
Avoid defining `Hash#deep_merge` and `#deep_merge!`

### LINKED ISSUE #314: avoid Hash monkey patch when loading DeepMerge
followup to #120

`DeepMerge` can be loaded in 3 different ways:
1. only the `core` library
2. the `core` library + the default `Hash` monkey patch
3. the `core` library + an alternative `Hash` monkey patch that does not conflict with `ActiveSupport`

`Config` is currently using the [2nd variant](https://github.com/danielsdeleo/deep_merge/blob/master/lib/deep_merge.rb):
https://github.com/rubyconfig/config/blob/9c35a5d579e548b2c439cd4defb247baf51dbb64/lib/config.rb#L9

However, `Config` does not actually use the monkey patch, so it should switch to the 1st variant:
```ruby
require 'deep_merge/core'
```

This is conceptually cleaner and prevents the conflict with `ActiveSupport`.

jonathanhefner:
`config` uses `DeepMerge.deep_merge!` instead of `Hash#deep_merge!`, so monkey patching `Hash` is unnecessary.  Furthermore, DeepMerge's `Hash` monkey patch is not compatible with Rails 7.1 (see [rails#49457][]).

This commit changes `require 'deep_merge'` to `require 'deep_merge/core'` so that DeepMerge's `Hash` monkey patch is no longer loaded.  Users who rely on the monkey patch can load it manually via `require 'deep_merge/deep_merge_hash'`.

Closes #314.

[rails#49457]: https://github.com/rails/rails/issues/49457


pkuczynski:
@cjlarose what do you think?

jonathanhefner:
> Could you please update the CHANGELOG as well?

Done!

pkuczynski:
@jonathanhefner I extended our test suite for Rails 7.x in #344, but test fails with a `can't modify frozen Array` error. We need this to verify your proposed solution. As I am not developing in Ruby anymore for several years, I am not able to fix it quickly. Would you mind having a look at #344?

pkuczynski:
#344 got finally merged so we can verify this PR agains Rails 7.x. Let's see if it works.

pkuczynski:
Looks it works! Merging and releasing...

pkuczynski (approved):

PR #350, opened 2024-02-06
Bump actions/checkout and actions/cache

m-nakamura145:
GitHub is [planning to upgrade to Node 20](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/). Versions prior to actions/checkout v3 and actions/cache v3  use an outdated version of node, so we will upgrade to v4, where Node 20 is the default. see [actions/checkout v4](https://github.com/actions/checkout/releases/tag/v4.0.0), [actions/cache v4](https://github.com/actions/cache/releases/tag/v4.0.0)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
           - truffleruby
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


pkuczynski (approved):
Sure, why not! Thanks!

PR #351, opened 2024-02-13
Prevent name collision with private methods from ancestors

vadtel:
I caught error when accessing an option named `lambda`
```
>> config['lambda']
# ArgumentError: tried to create Proc object without a block
from bundle/ruby/2.6.0/gems/config-2.2.3/lib/config/options.rb:161:in `lambda'
```
The methods used to access options are always public, so there is no need to use send. Using send can cause problems because it can call private and protected methods of the base classes. I replaced send with public_send. This will avoid the problems caused by send calling private and protected methods.

### DIFF
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -112,7 +112,7 @@ module Config
     end
 
     # Some keywords that don't play nicely with OpenStruct
-    SETTINGS_RESERVED_NAMES = %w[select collect test count zip min max table].freeze
+    SETTINGS_RESERVED_NAMES = %w[select collect test count zip min max exit! table].freeze
 
     # Some keywords that don't play nicely with Rails 7.*
     RAILS_RESERVED_NAMES = %w[maximum minimum].freeze
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -122,7 +122,7 @@ module Config
     def [](param)
       return super if SETTINGS_RESERVED_NAMES.include?(param)
       return super if RAILS_RESERVED_NAMES.include?(param)
-      send("#{param}")
+      public_send("#{param}")
     end
 
     def []=(param, value)
--- a/lib/config/options.rb
+++ b/lib/config/options.rb
@@ -112,7 +112,7 @@ module Config
     end
 
     # Some keywords that don't play nicely with OpenStruct
-    SETTINGS_RESERVED_NAMES = %w[select collect test count zip min max exit! table].freeze
+    SETTINGS_RESERVED_NAMES = %w[select collect test count zip min max table].freeze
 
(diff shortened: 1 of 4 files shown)


lib/config/options.rb
```diff
@@ -122,7 +122,7 @@ def merge!(hash)
     def [](param)
       return super if SETTINGS_RESERVED_NAMES.include?(param)
       return super if RAILS_RESERVED_NAMES.include?(param)
-      send("#{param}")
+      public_send("#{param}")
```
cjlarose, line 125: The reason I understand that this change works is because `lambda` and `proc` are both private methods (presumably of `Kernel`). Does that imply that we can remove some things out of `SETTINGS_RESERVED_NAMES` safely (like `exit!`)?

lib/config/options.rb
```diff
@@ -122,7 +122,7 @@ def merge!(hash)
...
     end
 
     def []=(param, value)
```
cjlarose, line 128: Any thoughts on whether or not it makes sense to use `public_send` in `[]=` as well?

vadtel: I think it doesnt make sense. Private setter doesnt make sense
And I dont know any private method like "\/.+=\" in the chain of inheritance
`Config::Options.ancestors.any? { |c| c == ::Class ? c.private_instance_methods.any? {|m| m[/.+=/]} : false }`

lib/config/options.rb
vadtel: Yes we can
I will do this

pkuczynski:
@cjlarose are you happy with the change? I would like to merge it and release it as `5.2.0`

pkuczynski:
@vadtel tests are failing for Ruby 3.0. Would you mind having a look?

vadtel:
> @vadtel tests are failing for Ruby 3.0. Would you mind having a look?

It depends on the `ostruct` version
We don't create singleton methods for fields with exclamation marks (bangs) how return `respond_to?(name, true)` - true
The method_missing method only returns nil in one version of Ruby
https://github.com/ruby/ostruct/commit/69cbbefd997f74b99b5135db8b9ef1ba97a31b43
May be we need to return "exit!" to constant?

pkuczynski:
> May be we need to return "exit!" to constant?

Yeah, possibly this would be the easiest solution. @cjlarose wdyt?

pkuczynski:
Waiting for the feedback from @cjlarose before we merge it in and release it. Hopefully soon...

pkuczynski (approved):

cjlarose (approved):

cjlarose:
Released with version 5.2.0

PR #353, opened 2024-03-06
Block reload config files in Railtie if they already uploaded

### LINKED ISSUE #352: Allow additional settings files to be specified when using Rails integration
In Rails projects, `config/initializers/config.rb` can be used to configure this gem, but the [Railtie](https://github.com/rubyconfig/config/blob/96132bab0f867795bc24d43532eb0ce2f852884e/lib/config/integrations/rails/railtie.rb) performs the actual first load of the settings files. The Railtie doesn't offer a way to specify additional settings sources (for example, multiple YAML files).

It's possible to specify additional settings files today if users use a separate initializer, but this isn't always convenient because it essentially means that settings are loaded twice (and both times must satisfy the configured schema or contract).

Related: https://github.com/rubyconfig/config/pull/142

vovanmozg:
Fixes https://github.com/rubyconfig/config/issues/352


spec/config_spec.rb
```diff
@@ -151,6 +151,46 @@
     expect(Settings.size).to eq(2)
   end
 
+  describe "load_and_set_settings in the Rails context" do
```
cjlarose, line 154: Awesome tests!

They seem to be failing on CI though. There might be something going on with test execution order as well since locally sometimes I get 3 failures, and other times just one. I'm able to reproduce the failures locally on ruby 3.1.4

```sh
bundle exec appraisal rails-6.1 rspec
```

vovanmozg: Hi! 

I achieved stable running of tests locally. `Config::Options` test cleared ENV. So `rails runner` executed with empty ENV and this led to an error starting rails (Gem not found).

vovanmozg:
@cjlarose, could you check tests?

pkuczynski:
@Nuzair46 what do you think about this one?

Nuzair46:
@pkuczynski @vovanmozg  the failing tests seems to be an issue with Rails 6.1 not loading logger properly.
i was able to fix it by doing this
Before:
```rb
`#{Rails.root.join("bin", "rails")} runner 'print "\#{Settings.default}|\#{Settings.custom}"'`
```
After:
```rb
`RUBYOPT="-r logger" #{Rails.root.join("bin", "rails")} runner 'print "\#{Settings.default}|\#{Settings.custom}"'`
```
This makes ruby to load the logger before anything runs.

vovanmozg:
> This PR will be superseded by #382 @vovanmozg Please check that PR if it satisfies your changes

Thank you! 
Looks good to me

Nuzair46:
Will be done in #382

PR #354, opened 2024-03-08
Allow arrays to be passed through env variables

dominh:
I think i'd like to have it in my projects in which i'm using Config gem

### DIFF
--- a/lib/config/sources/env_source.rb
+++ b/lib/config/sources/env_source.rb
@@ -6,17 +6,20 @@ module Config
       attr_reader :separator
       attr_reader :converter
       attr_reader :parse_values
+      attr_reader :parse_arrays
 
       def initialize(env,
                      prefix: Config.env_prefix || Config.const_name,
                      separator: Config.env_separator,
                      converter: Config.env_converter,
-                     parse_values: Config.env_parse_values)
+                     parse_values: Config.env_parse_values,
+                     parse_arrays: Config.env_parse_arrays)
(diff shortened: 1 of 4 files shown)


lib/config/sources/env_source.rb
```diff
@@ -52,10 +52,28 @@ def load
           leaf[keys.last] = parse_values ? __value(value) : value
         end
 
-        hash
+        convert_hashes_to_arrays(hash)
```
cjlarose, line 55: These changes look good!

The only change I'd like to see if a new option that enables this behavior (similar to `env_parse_values`) and to set it to `false` by default so it's not likely to break anyone's existing config.

```ruby
Config.setup do |config|
  config.use_env = true
  config.env_parse_arrays = true
end
```

spec/config_env_spec.rb
```diff
@@ -22,6 +22,7 @@
       Config.env_separator        = '.'
       Config.env_converter        = :downcase
       Config.env_parse_values     = true
+      Config.env_parse_arrays     = true
```
dominh, line 25: Why this entire block is here? It actually makes specs in this file being unable to test gem's defaults

cjlarose: Yeah, I agree that's weird. It'd be better if each test was able to rely on having the defaults set and then be explicit about when they should be different

cjlarose (approved):

Issue #68, opened 2014-03-07
Integration with encrypted Rails credentials


smith:
Rails 4.1 introduces a built-in convention for having a secrets.yml with secrets in it: http://edgeguides.rubyonrails.org/4_1_release_notes.html#config-secrets-yml

It would be cool if this gem did some of these things:
- Aliased `Rails.application.secrets` to `Settings.secrets`
- Shimmed the secrets.yml functionality in for Rails < 4.1
- Allowed you to have a secrets.local.yml or config/secrets/production.yml that behave the same way as the existing conventions do for non-secrets files

I think this gem is a better overall solution to managing configuration, but if Rails is making this a thing, it would make sense to go along with it and make this gem work nicely with the new stuff.

pkuczynski:
@rdubya would you like to have a look at this one?

rdubya:
Rails 5 gets away from using secrets and now has a `credentials` file. It would be cool to have similar functionality built into this gem since it provides a much more flexible way of managing environment settings.

pkuczynski:
I think we could add alias `Rails.application.secrets` to `Settings.secrets` if anybody is interested in this. However, I would keep the way  `secrets.yml` work meaning with environments groups.

@smith @Darkside73 are you still interested in this?

smtlaissezfaire:
FYI This bug caused me so much pain...

the config gem eager loads credentials, causing any overrides to not be respected in config/application.rb

pkuczynski:
@smtlaissezfaire I would welcome a PR fixing this. Would like to have a look? I no longer work with Rails for few years now so it would be hard for me to get back to it...

noxasch:
Just sharing my workaround to work with `Rails.application.credentials`. Not sure if this the optimum, but this monkey patching works for me to utilize Setting and Rails secret management.

```rb
# config/initializers/config.rb
module Config
  def self.load_and_set_settings(*sources)
    name = Config.const_name
    Object.send(:remove_const, name) if Object.const_defined?(name)
    Object.const_set(name, Config.load_files(sources))
  end

  def self.load_files(*sources)
    config = Options.new

    # add settings sources
    [sources].flatten.compact.each do |source|
      config.add_source!(source)
    end

    config.add_source!(Sources::EnvSource.new(ENV)) if Config.use_env
    load_credentials(config) # call Settings override
    config.load!
    config
  end
 
  # override with crendentials
  def self.load_credentials(config)
    if Rails.env.production? || Rails.env.staging?
[...]

PR #355, opened 2024-03-08
feat: add rails credentials support

noxasch:
Added rails credentials support with config flag addressing #68


noxasch:
I'm currently having trouble with different rails version in the test, any clue ?

Screenshot from 2024-03-14 13-39-57

cjlarose:
Not 100% sure what's going on with the tests on CI. Tests pass for me locally. I suspect it has _something_ to do with Rails 7.0 or 7.1 because we don't run tests for those Rails versions when running the test suite for Ruby 2.7, jruby, or truffleruby.

I merged in a change that address the deprecation warnings for `fixture_path` in Rails 7.1 to help clear up the test output. Hoping that will help clear things up 🤞

lib/config.rb
```diff
@@ -47,6 +48,14 @@ def self.load_files(*sources)
 
     config.add_source!(Sources::EnvSource.new(ENV)) if Config.use_env
 
+    if defined?(::Rails::Railtie) && Config.use_rails_credentials
+      if Rails.application.credentials.to_h.present?
```
cjlarose, line 52: I think when some of the tests run, even on Rails 7.1, `Rails.application.credentials.to_h` will be an empty `Hash`, which will cause us to go into the `else` branch, which will access `Rails.application.secrets`, which causes a deprecation warning. Maybe we can use `respond_to` or something to avoid this from happening

```suggestion
      if Rails.application.respond_to?(:credentials)
```

This _might_ or might not be related to test failures. I think we do try to raise on deprecation warnings in the test suite.

noxasch: Thanks @cjlarose,  I did update the reset method in my stash. I'll see what else I can do. My local is using Ruby 3.2.2 with Rails 7.1 and the test pass as well on my local.

noxasch:
Update: 

Rebased to latest master 

I look around and found the solution for rails 7.1 fail ted. and fix issue with rails 7.1. Rails 7.1 and above need to use `ActiveSupport::EncryptedConfiguration` config method.

All test pass now.

lib/config.rb
```diff
@@ -48,6 +49,14 @@ def self.load_files(*sources)
 
     config.add_source!(Sources::EnvSource.new(ENV)) if Config.use_env
 
+    if defined?(::Rails::Railtie) && Config.use_rails_credentials
+      if Rails.version < '7.1'
```
cjlarose, line 53: Because the version numbers are represented as strings here, this comparison can lead to unexpected results. For example, if `Rails.version` is `10.0.0`, then `'10.0.0' < '7.1' == true`

To compare version numbers correctly, I think we need to either use something like

```ruby
if [Rails::VERSION::MAJOR, Rails::VERSION::MINOR] < [7, 1]
```

or

```ruby
if Gem::Version.new(Rails.version) < Gem::Version.new('7.1')
```

lib/config.rb
```diff
@@ -48,6 +49,14 @@ def self.load_files(*sources)
...
+        config.add_source!(Sources::HashSource.new(secret: Rails.application.secrets.to_h.deep_stringify_keys))
```
cjlarose, line 54: Let's remove the nesting of the credentials under the top-level `secret` key. I think that'll be more useful as the default behavior in order to support merging with other config sources. For example, if my config.yaml file has

```yaml
aws:
  region: 'us-west-1'
```

And my Rails credentials has

```yaml
aws:
  secret_access_key: '123456'
```

...then I'd want to be able to access `Settings.aws`, which would contain both `region` and `secret_access_key`.

If someone _really_ wants to keep their secrets in a separate key, they can always turn off `use_rails_credentials` add it themselves in an initializer

```ruby
Settings.add_source!({ secret: Rails.application.credentials.config.deep_stringify_keys })
Settings.reload!
```

lib/config.rb
```diff
@@ -48,6 +49,14 @@ def self.load_files(*sources)
...
+      else
+        config.add_source!(Sources::HashSource.new(secret: Rails.application.credentials.config))
```
cjlarose, line 56: I think `Rails.application.credentials.config` returns a hash with symbol keys. I think for merging to work correctly, we need to `.deep_stringify_keys` here.

lib/config.rb
```diff
@@ -48,6 +49,14 @@ def self.load_files(*sources)
 
     config.add_source!(Sources::EnvSource.new(ENV)) if Config.use_env
```
cjlarose, line 50: Let's load credentials _before_ the environment. My expectation is usually that env vars should "win" against every other configuration source.

spec/app/rails_7.1/config/credentials/test.yml.enc
```diff
@@ -0,0 +1 @@
+qU0cAsjfKz4lnoJPU36JuM7Yh3qm6B2YV7LhqJROKCu4AsOW0AfFY9FM+aTeRsZOIdIBHyyargCI1xmq3N5o4rdVZRXxWIt2PD93xZPlcMrlAb645m8hyni1cW4=--dPokgPsIyoKhzMzD--nb/fnuH1FaTNT5iF8J3TfQ==
```
cjlarose, line 1: I tried looking at these values locally with

```sh
cd spec/app/rails_7.1 
RAILS_MASTER_KEY='0e29551e5c31acf7c769d64397af54e4' bundle exec rails credentials:edit
```

but I got an error

```
Editing config/credentials.yml.enc...
Couldn't decrypt config/credentials.yml.enc. Perhaps you passed the wrong key?
```

...but it seems to work on CI, so it's a little weird. Any idea how I can look at these values locally?

cjlarose (changes requested):
Awesome work! Requested a few changes

lib/config.rb
noxasch: Awesome, agree on this.

spec/app/rails_7.1/config/credentials/test.yml.enc
noxasch: @cjlarose just remove the single quotes

noxasch:
Update:

1. Addressed the changes and refactor the code.
2. Update the crendentails to include the aws patter below to be more realistic
```yml
aws:
  secret_access_key: '123456'
```

Apparently we don't need to check for rails version, just need to require the master_key in test environment and for rails5.2 untiil 6.1 need test.key and master.key, otherwise it somehow ignore the RAILS_MASTER_KEY in environment variable

noxasch:
@cjlarose let me know if there is anything else require.

nickpoorman:
@cjlarose @noxasch can we please get this merged? Thanks!

pkuczynski:
@cjlarose since you reviewed in the past, would you mind doing a re-review? Seems that all your feedback was addressed?

@noxasch would you mind adding a CHANGELOG entry?

lib/config.rb
```diff
@@ -47,6 +48,15 @@ def self.load_files(*sources)
       config.add_source!(source)
     end
 
+    # load rails crendentials
+    if defined?(::Rails::Railtie) && Config.use_rails_credentials
+      if Rails.application.credentials.respond_to?(:credentials)
```
toncid, line 53: Should this be `Rails.application.respond_to?(:credentials)`? Or just do Rails version detection?

noxasch: sure. Honestly i lose memory on why I opt to `Rails.application.credentials.respond_to?(:credentials)` when i created this PR, but if `Rails.application.respond_to?(:credentials)` works, then why not.

noxasch:
@pkuczynski I see all the test pass already, guess you fix all the test ?

pkuczynski:
Yes, was able to fix the tests thanks to @Nuzair46 help. Nowe you need to resolve conflicts and we should be good to go @noxasch

noxasch:
@Nuzair46 not sure what you mean by that. What this does it will read rails credentials merge it with existing config from your `settings.yml` and if you set `use_env`, then your setting that use ENV will take preceding over other including in the credentials.

Nuzair46:
@noxasch I was wondering if this pr enables evaluating the credentials file.
So previously without credentials in rails, we could write the file like this
```yml
api_key: '12345'
api_key_env: ENV['API_KEY'] # this will be evaluated and load the variable provided in the env
```
But after Rails credentials, this won't work anymore. And I have seen people wanting this feature like being able to have some env variable load from vault or other env managers. I thought this change to config gem will also make this possible, but it seems like it doesn't. 
This is a test done on this branch to see if it works.
```rb
tokyo(dev)> Rails.application.credentials.api_key
=> "12345"
tokyo(dev)> Rails.application.credentials.api_key_env
=> "<%= ENV[\"API_KEY\"] %>"
tokyo(dev)> exit
╭─Red@Red ~/Miqor/Tokyo ‹main●› 
╰─$ rails c
Loading development environment (Rails 8.0.2)
tokyo(dev)> Settings.api_key
=> "12345"
tokyo(dev)> Settings.api_key_env
=> "<%= ENV[\"API_KEY\"] %>"
tokyo(dev)> exit
╭─Red@Red ~/Miqor/Tokyo ‹main●› 
╰─$ API_KEY=123 rails c
Loading development environment (Rails 8.0.2)
tokyo(dev)> Settings.api_key_env
=> "<%= ENV[\"API_KEY\"] %>"
tokyo(dev)> Settings.api_key
=> "12345"
tokyo(dev)> Rails.application.credentials.api_key_env
=> "<%= ENV[\"API_KEY\"] %>"
```

Nuzair46 (dismissed):

noxasch:
@Nuzair46 Understand. I supposed that is how `config.add_source!` work here,if we want to achieve ENV to override any settings, then need to change something i guess.

Nuzair46:
> @Nuzair46 Understand. I supposed that is how it previously work that it only change if the setting is nil, if we want to achieve this should be a different change require there.

Yea it will require us to evaluate the env and then attach it. But I think it might be out of scope for this repo.

lib/config.rb
Nuzair46: @noxasch @pkuczynski 
This is a bit tricky. `credentials` and `secrets` are both available on `rails 5.2 - 7.1`. In `rails >= 7.2` `secrets` are not available. So, this condition will always return true whether the app uses `credentials` or not and the else part would not run.
Here the condition should check for Rails version and load both credentials and secrets for Rails versions that have both and only credentials for `rails >= 7.2`. I think credentials came with `rails 5.2` and we only support `rails >= 6.0`, so we don't need to load just secrets.

README.md
```diff
@@ -290,6 +290,7 @@ After installing `Config` in Rails, you will find automatically generated file t
 * `evaluate_erb_in_yaml` - evaluate ERB in YAML config files. Set to false if the config file contains ERB that should not be evaluated at load time. Default: `true`
 * `file_name` - name of the file to store general keys accessible in all environments. Default: `'settings'` - located at `config/settings.yml`
 * `dir_name` - name of the directory to store environment-specific files. Default: `'settings'` - located at `config/settings/`
+* `use_rails_credentials` - evaluate Rails credentials if loaded with `RAILS_MASTER_KEY` or `config/master.key`. Can be access using `Settings.secret.secret_key_base`. Rails credetials will override other settings as it will loaded last.
```
Nuzair46, line 293: Would be nice to also mention it also loads both `credentials` and `secrets` in case of `rails < 7.2` and how the load order we do might affect it (depending on the `config.add_source!` in `lib/config.rb`

Nuzair46:
Hey @noxasch, Ive added some comments. Please check. rest looks good.

Nuzair46 (approved):
LGTM

lib/config.rb
Nuzair46: solvede

lib/config.rb
Nuzair46: solved

Nuzair46:
@pkuczynski can we get this merged? it LGTM from my side.

PR #356, opened 2024-03-15
Configuring the environment name in Rails

neilwilliams:
I've finally found some time to raise a new PR for this. I originally raised it as https://github.com/rubyconfig/config/pull/290.

This small new configuration option allows us to specify the name for the rails environment, rather than being limited to using 'development', 'test' and 'production'.

We have multiple rails 'production' environments, so this is needed to support that.

For example, when this new config is enabled (it defaults to how it works now), we would need to set the new environment variable:
```
ENVIRONMENT=preprod
```
or
```
ENVIRONMENT=perf
```

And have an associated config for it in ./config:. e.g.
```
./config
  settings.yml
  settings/
    preprod.yml
    perf.yml
```

@cjlarose FYI - hopefully we can get this in this time, and we can stop referencing a copy of this gem from our private gem repository 😄


cjlarose (approved):

cjlarose:
Previously, we wanted a test for the Railtie, but that ended up being a big hassle. I think we're gonna figure it out in https://github.com/rubyconfig/config/pull/353 though, so I'm comfortable merging this in as-is and we can add a test later

cjlarose:
Released in version 5.4.0

neilwilliams:
Thanks @cjlarose !

PR #363, opened 2024-09-06
Fix warning: ostruct was loaded from the standard library

### LINKED ISSUE #365: Add "ostruct" to gemspec/Gemfile
Hi,

Can you add "ostruct" to gemspec/Gemfile  to silence this warning and handle future ruby bump issues.

```sh
.../gems/ruby-3.3.5/gems/config-5.5.1/lib/config/options.rb:1: warning: .../rubies/ruby-3.3.5/lib/ruby/3.3.0/ostruct.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
```

taketo1113:
It fixes a warning of loading ostruct gem from standard library with Ruby 3.3.5 and 3.4+.

The warning message is following:
```
/path/config/lib/config.rb:1: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
```

- ruby: 3.3.5
- config gem: 5.5.1

### Related Links
- https://github.com/ruby/ruby/pull/10428
- https://bugs.ruby-lang.org/issues/20309

### Additional information
This warning category is `performance`.
It may be better to reduce the dependency of ostruct.

> OpenStruct use is discouraged for performance reasons

https://github.com/ruby/ostruct/pull/56

Closes #365

### DIFF
--- a/config.gemspec
+++ b/config.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.6.0'
 
   s.add_dependency 'deep_merge', '~> 1.2', '>= 1.2.1'
+  s.add_dependency 'ostruct'
 
   s.add_development_dependency 'dry-validation', *Config::DryValidationRequirements::VERSIONS
   s.add_development_dependency 'rake', '~> 12.0', '>= 12.0.0'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Fix warning: ostruct was loaded from the standard library ([#363](https://github.com/rubyconfig/config/pull/363))
+
 ## 5.5.1
 
 ### Documentation


config.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.6.0'
 
   s.add_dependency 'deep_merge', '~> 1.2', '>= 1.2.1'
+  s.add_dependency 'ostruct' if RUBY_VERSION >= '3.3.5'
```
Fryguy, line 31: I don't think you can do conditionals like this in a gemspec, because when you gem package it evaluates this at package time.  That is, if it's packaged on < 3.3.5 it won't exist, and if it's packaged on 3.3.5+ it will exist, which means it's non-deterministic.

Fryguy:
> It may be better to reduce the dependency of ostruct.

ostruct is really a core component of this gem, because that's what provides the dot access to all the settings.

What makes ostruct a performance issue is when you have to create a lot of them only to access their settings a small amount of times (high create-low read/write), because ostruct creates a lot of method accessors on the singleton object under the covers.  The overhead of creating those method accessors outweighs the time to access the underlying value. A popular way to do it differently is to use something like method missing, but that flips the performance issue.  If you have to access a particular value a lot of times on a cached/singleton object, the overhead of the extra method_missing dispatch outweighs the method creation

In general, Config falls into the latter category.  When you start your app, it creates all the ostructs, and then that's done, generally for the lifetime of the app, and you can access as many times as your app needs.  I would guess most app have a very high read rate if they have do it in every request, for example.

So, if we wanted to drop ostruct we could easily replace ostruct with something like method_missing, but that would actually likely introduce a new performance problem on the high-read side.  Otherwise we'd have to duplicate ostruct's ability to create method accessors, which probably wouldn't be too bad, but feels like overhead when there's a gem that does that.

config.gemspec
taketo1113: @Fryguy  Thank you for a comment.
I removed condition.

taketo1113:
@Fryguy As you say, there is no need to replace ostruct.

I created PoC to compare ostruct with method_missing.
The benchmark result is below.
It comfirm a new performance problem on the high-read side with `MethodMissingClass (assign/read)`.

```
                                            user     system      total        real
OpenStruct.new (without args)           0.001136   0.000023   0.001159 (  0.001158)
MethodMissingClass.new (without args)   0.000699   0.000004   0.000703 (  0.000702)
OpenStruct (assign)                     0.000655   0.000001   0.000656 (  0.000657)
MethodMissingClass (assign)             0.002662   0.000009   0.002671 (  0.002674)
OpenStruct (read)                       0.000727   0.000000   0.000727 (  0.000727)
MethodMissingClass (read)               0.002228   0.000021   0.002249 (  0.002248)
```

<details><summary>PoC & Benchmark</summary>

### PoC & Benchmark code
```ruby
# benchmark.rb
require 'benchmark'
require 'ostruct'

puts RUBY_DESCRIPTION

n = 10_000
puts "times: #{n}"

# OpenStruct
ostruct = OpenStruct.new

# method_missing
class MethodMissingClass
  def method_missing(name, *value)
    self.class.class_eval do
      define_method "#{name}" do |*value|
        if name.end_with?('=')
          instance_variable_set("@#{name.to_s.chop}", *value)
        else
          instance_variable_get("@#{name}")
        end
      end
    end
    send(name, *value)
  end
end
method_missing = MethodMissingClass.new

# Benchmark
Benchmark.bmbm do |x|
  x.report("OpenStruct.new (without args)") { n.times { OpenStruct.new } }
  x.report("MethodMissingClass.new (without args)") { n.times { MethodMissingClass.new } }
  x.report("OpenStruct (assign)") { n.times { ostruct.a = 1 } }
  x.report("MethodMissingClass (assign)") { n.times { method_missing.a = 1 } }
  x.report("OpenStruct (read)") { n.times { ostruct.a } }
  x.report("MethodMissingClass (read)") { n.times { method_missing.a } }
end
```

### Benchmark Result:
```
$ ruby --yjit benchmark_method_missing.rb
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) +YJIT [arm64-darwin23]
times: 10000
Rehearsal -------------------------------------------------------------------------
OpenStruct.new (without args)           0.001591   0.000190   0.001781 (  0.001780)
MethodMissingClass.new (without args)   0.000912   0.000011   0.000923 (  0.000923)
OpenStruct (assign)                     0.000753   0.000028   0.000781 (  0.000781)
MethodMissingClass (assign)             0.003194   0.000066   0.003260 (  0.003263)
OpenStruct (read)                       0.000826   0.000021   0.000847 (  0.000848)
MethodMissingClass (read)               0.002592   0.000057   0.002649 (  0.002652)
---------------------------------------------------------------- total: 0.010241sec

                                            user     system      total        real
OpenStruct.new (without args)           0.001136   0.000023   0.001159 (  0.001158)
MethodMissingClass.new (without args)   0.000699   0.000004   0.000703 (  0.000702)
OpenStruct (assign)                     0.000655   0.000001   0.000656 (  0.000657)
MethodMissingClass (assign)             0.002662   0.000009   0.002671 (  0.002674)
OpenStruct (read)                       0.000727   0.000000   0.000727 (  0.000727)
MethodMissingClass (read)               0.002228   0.000021   0.002249 (  0.002248)
```
</details>

pkuczynski:
@taketo1113 updating changelog would be useful

taketo1113:
@pkuczynski Currently, CHANGELOG does not have a section of Unreleased.
Could I add a section of `Unreleased` at CHANGELOG?
https://github.com/rubyconfig/config/blob/86c47f25f3ed1eb29827bfcf6f1a30d811a964b7/CHANGELOG.md

pkuczynski:
> @pkuczynski Currently, CHANGELOG does not have a section of Unreleased. Could I add a section of `Unreleased` at CHANGELOG? https://github.com/rubyconfig/config/blob/86c47f25f3ed1eb29827bfcf6f1a30d811a964b7/CHANGELOG.md

Yes please...

taketo1113:
@pkuczynski I updated CHANGELOG.

pkuczynski:
@Fryguy @cjlarose are you happy with those changes?

Fryguy (approved):

jsugarman (approved):

jsugarman:
This [issue](https://github.com/rubyconfig/config/issues/365) can be closed once merged 👍

gstokkink:
@Fryguy can we get a release for this fix? 😄

Fryguy:
I'd love to but I'm not a gem owner. @pkuczynski can do it once this is merged ☺️

gstokkink:
Ah, okay! @pkuczynski any chance of a release soon for this fix? Everyone upgrading to Ruby 3.3.5 will run into this warning, and for CIs that forbid deprecation warnings, this is kind of an issue 😋

pkuczynski (approved):

pkuczynski:
Sure! Done.

PR #364, opened 2024-09-16
Add ostruct as dependency to gemspec

gstokkink:
This should silence warnings like:

```
.../ruby/3.3.0/gems/config-5.5.1/lib/config/options.rb:1: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.00:03
You can add ostruct to your Gemfile or gemspec to silence this warning.
```

See https://github.com/ruby/ruby/pull/10428 for more information.


gstokkink:
Closing in favour of https://github.com/rubyconfig/config/pull/363.

PR #367, opened 2024-11-13
feat: make `deep_merge` a soft-requirement

zzak:
The goal here is to remove deep_merge and rely entirely on ActiveSupport's `Hash#deep_merge` implementation.

This does mean that if you want to keep the existing behavior, or rely on DeepMerge's specific options, you need to add the following to your Gemfile:

```ruby
gem 'deep_merge', '~> 1.2', '>= 1.2.1'
```

---

@pkuczynski Opening this as a proof-of-concept, would you be willing to take on this type of change?

I decided not to touch the config methods, even though they are no-op when `deep_merge` gem isn't included. Wanted to get feedback first, and I figure the warning is enough.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       COVERAGE_RUBY_VERSION: 2.6
-      BUNDLE_PATH: vendor/bundle
+      BUNDLE_PATH: ${{ github.workspace }}/vendor/bundle
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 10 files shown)


pkuczynski:
@Nuzair46 what do you think about this one?

lib/config.rb
```diff
@@ -6,7 +6,18 @@
 require 'config/sources/hash_source'
 require 'config/sources/env_source'
 require 'config/validation/schema'
-require 'deep_merge/core'
+
+begin
+  gem 'deep_merge', '~> 1.2', '>= 1.2.1'
+  require 'deep_merge/core'
+rescue LoadError
```
Nuzair46, line 13: I think here like @pkuczynski said, it is better to raise an error if deep_merge is not available in a non Rails codebase. Since this gem is can be used in any ruby codebase, it is better to warn for Rails and raise error for non rails app if it is not available,

PR #371, opened 2025-07-08
ci: bump rails to the latest versions and add missing deps

pkuczynski:


Nuzair46:
@pkuczynski jruby 10 has an issue with loading constants and files https://github.com/jruby/jruby/issues/8777
for now I think this can be ignored or use jruby 9 if testing against jruby is needed.

PR #387, opened 2026-01-28
Apply rubocop and enable in CI

Nuzair46:
This PR applies rubocop rules (mostly auto fixes) and enables rubocop check in the CI


lib/config/options.rb
```diff
@@ -70,25 +70,29 @@ def reload_from_files(*files)
...
     end
 
-    alias :to_h :to_hash
+    alias to_h to_hash
 
     def each(*args, &block)
       marshal_dump.each(*args, &block)
     end
 
+    def each_key(*args, &block)
```
Nuzair46, line 90: Added this new method for a more hash like `Config`

lib/config/sources/env_source.rb
```diff
@@ -25,27 +21,27 @@ def initialize(env,
       def load
         return {} if @env.nil? || @env.empty?
 
-        hash = Hash.new
+        hash = {}
```
Nuzair46, line 24: Since there are no defaults set here, we can just use literal

